PR #3, opened 2014-12-18
Provide an interface for the random string generator

pfrenssen:
The random string generator `\Drupal\Component\Utility\Random` is currently hardcoded in the project. It would be better if this conformed to an interface so it can be properly injected. This would allow us to override it, and it allow the methods to be autocompleted by IDEs.

### DIFF
--- a/src/Drupal/Driver/DrushDriver.php
+++ b/src/Drupal/Driver/DrushDriver.php
@@ -3,6 +3,7 @@
 namespace Drupal\Driver;
 
 use Drupal\Component\Utility\Random;
+use Drupal\Component\Utility\RandomInterface;
 use Drupal\Driver\Exception\BootstrapException;
 
 use Symfony\Component\Process\Process;
@@ -41,7 +42,7 @@ class DrushDriver extends BaseDriver {
   /**
    * Random generator.
    *
-   * @var \Drupal\Component\Utility\Random
+   * @var \Drupal\Component\Utility\RandomInterface
    */
   private $random;
 
@@ -61,12 +62,12 @@ class DrushDriver extends BaseDriver {
    *   The root path of the Drupal install. This is an alternative to using aliases.
    * @param string $binary
    *   The path to the drush binary.
-   * @param \Drupal\Component\Utility\Random $random
+   * @param \Drupal\Component\Utility\RandomInterface $random
    *   Random generator.
    *
    * @throws \Drupal\Driver\Exception\BootstrapException
    */
-  public function __construct($alias = NULL, $root_path = NULL, $binary = 'drush', Random $random = NULL) {
+  public function __construct($alias = NULL, $root_path = NULL, $binary = 'drush', RandomInterface $random = NULL) {
     if (!empty($alias)) {
       // Trim off the '@' symbol if it has been added.
       $alias = ltrim($alias, '@');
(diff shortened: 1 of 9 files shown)


jhedstrom:
Sorry for the delayed response.

The reason there isn't an interface here already is that the random generator has been plucked directly from Drupal 8 (I was hoping the components there would eventually be available individually via `composer.json`).

Ideally, I'd like to see Drupal 8 adopt an interface approach, but in the meantime, I think this approach makes sense to do here.

Somebody on twitter mentioned using https://github.com/rchouinard/rych-random as a random generator, I wonder if that would be possible still with this approach?

pfrenssen:
I have used rych-random in a previous project but we eventually abandoned it for a self rolled solution since it is rather basic. What is very nice about it is that it scans your system for available random number generators and will use the best available.

We decided to abandon it because in testing it is very interesting to be able to retrieve valid data as well as invalid data, and also to be able to extend it so you can test things like URLs, email addresses etc.

This is the interface we used, it served us pretty well (ref http://cgit.drupalcode.org/paddle_selenium_tests/tree/testdataprovider/Kanooh/TestDataProvider/TestDataProviderInterface.php?h=7.x-1.x):

```
<?php


/**
 * @file
 * Contains \Kanooh\TestDataProvider\TestDataProviderInterface.
 */

namespace Kanooh\TestDataProvider;

/**
 * Interface for TestDataProvider classes.
 */
interface TestDataProviderInterface
{
    /**
     * Returns an array of valid test data.
     *
     * @return array
     *   An array of valid test data.
     */
    public function getValidDataSet();

    /**
     * Returns an array of invalid test data.
     *
     * @return array
     *   An array of invalid test data.
     */
    public function getInvalidDataSet();

    /**
     * Returns a random value from the valid data set.
     *
     * @return mixed
     *   A valid value.
     */
    public function getValidValue();

    /**
     * Returns a random value from the invalid data set.
     *
     * @return mixed
     *   An invalid value.
     */
    public function getInvalidValue();
}
```

This can be coupled with a base class that implements `getValidValue()` and `getInvalidValue()`.

Here is the full set of classes: http://cgit.drupalcode.org/paddle_selenium_tests/tree/testdataprovider/Kanooh/TestDataProvider?h=7.x-1.x

src/Drupal/Driver/DrushDriver.php
```diff
@@ -82,6 +83,7 @@ public function __construct($alias = NULL, $root_path = NULL, $binary = 'drush',
 
     $this->binary = $binary;
 
+    // @todo This should not be optional but injected by a factory method.
     if (!isset($random)) {
       $random = new Random();
```
jhedstrom: This won't actually work as written now, since `Random` won't be available without a `use` statement. I'd be happy to change this to using a factory method if you want to take a go at that.

jhedstrom:
I'd love to actually get an interface into Drupal core instead of just the Random class.

jhedstrom:
If my feedback above in the DrushDriver is resolved, I'm open to making this change as-is, and we can always follow-up in core.

jhedstrom:
We no longer directly hardcode the `Random` class, but rather require it via composer. I think the proper fix is to get an interface into core.

jhedstrom:
I've opened https://www.drupal.org/project/drupal/issues/2969060. Closing this out for now.

PR #8, opened 2015-02-11
Make module list work again

marcingy:


jhedstrom:
I don't thing `$this->getModuleList()` will work, since the methods live on the cores, and there isn't a wrapper method on the DrupalDriver.

It probably needs to be `$this->getCore()->getModuleList()`.

pfrenssen:
I want to help on this but Github is really not a good tool for collaborating on issues. On drupal.org I could simply post a new patch. What do I have to do here? Fork the fork?

pfrenssen:
Ok so I cannot fork Marc's fork. I have already forked jhedstrom/DrupalDriver and GitHub just redirects me to my already existing fork.

What I can do is add Marc's fork as a remote to mine, then fetch his repo and cherry-pick or merge that into a branch of my own. But then if I want to share the code I can only do that by making a new pull request which would mean we end up with two pull requests for the same thing.

So it's basically not possible to collaborate on tickets in GitHub without manually granting people permission.

jhedstrom:
Yeah, github is not the best for collaboration. I went ahead and merged this, and added a fix for `getCore()`.

PR #11, opened 2015-02-24
Create content in language with field translation.

dutchiexl:
You can add a language in the table hash
<pre>
 | title        | News feature: News item 1 |
 | body         | This is a news item       |
 | field_date   | 2018-02-25 07:45:45       |
 | language     | en                        |
</pre>

Every field that is translatable, will be set in this language

### DIFF
--- a/src/Drupal/Driver/Fields/Drupal7/AbstractHandler.php
+++ b/src/Drupal/Driver/Fields/Drupal7/AbstractHandler.php
@@ -10,6 +10,10 @@ namespace Drupal\Driver\Fields\Drupal7;
 use Drupal\Driver\Fields\FieldHandlerInterface;
 
 abstract class AbstractHandler implements FieldHandlerInterface {
+  /**
+   * @var string
+   */
+  protected $language = LANGUAGE_NONE;
 
   /**
    * @var
@@ -22,8 +26,10 @@ abstract class AbstractHandler implements FieldHandlerInterface {
    * @param $field_name
    * @return mixed
    */
-  public function __construct($entity_type, $field_name) {
+  public function __construct($entity_type, $field_name, $language) {
     $this->field_info = field_info_field($field_name);
+    if ($this->field_info['translatable']) {
+      $this->language = $language;
+    }
   }
-
 }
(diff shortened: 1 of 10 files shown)


jhedstrom:
Awesome, thanks!

ademarco:
@jhedstrom @dutchiexl : I'm afraid the proposed solution breaks in Drupal 8 since ```\Drupal\Driver\Cores\AbstractCore``` is supposed to be Drupal core agnostic, so in there we cannot refer to Drupal 7 specifics like the ```LANGUAGE_NONE``` constant, which is not available in Drupal 8.

With current master branch of both drupal extension and drupal driver, when running D8 tests, I get:

```
...
@d8
  Scenario: Create many nodes                                  # features/api.feature:49
    Given "page" content:                                      # Drupal\DrupalExtension\Context\DrupalContext::createNodes()
      | title    |
      | Page one |
      | Page two |
      Notice: Use of undefined constant LANGUAGE_NONE - assumed 'LANGUAGE_NONE' in vendor/drupal/drupal-driver/src/Drupal/Driver/Cores/AbstractCore.php line 42
    And "article" content:       
...
```

ademarco:
We could actually add the $entity object to the field handler constructor, so we delegate all core specific logic the handler implementations, so from:

```php
public function getFieldHandler($entity_type, $field_name, $language);
```
 We would have:

```php
public function getFieldHandler($entity, $entity_type, $field_name);
```

src/Drupal/Driver/Fields/Drupal7/AbstractHandler.php
```diff
@@ -22,8 +26,8 @@
    * @param $field_name
    * @return mixed
    */
-  public function __construct($entity_type, $field_name) {
+  public function __construct($entity_type, $field_name, $language) {
     $this->field_info = field_info_field($field_name);
+    $this->language = field_is_translatable('node', $this->field_info) ? $language : LANGUAGE_NONE;
```
ademarco: Field handlers are actually supposed to work with all type of entities, so any specific entity type reference (i.w. ```'node'``` in this case) should be removed, ```$entity_type``` should be used instead.

ademarco: We could actually use something similar to what core does for comments:

```php
$langcode = field_is_translatable('comment', $field) ? entity_language('comment', $comment) : LANGUAGE_NONE;
```

So:

```php
$langcode = field_is_translatable($entity_type, $this->field_info) ? entity_language($entity_type, $entity) : LANGUAGE_NONE;
```

dutchiexl: Yes, comments is where I got my mustard from. I can check out the entity_language if you like and refactor the update.

PR #19, opened 2015-02-27
Casted values to array.

dutchiexl:
Casted all handler field values variables to arrays, because most of the time they are not an array but a string.

### DIFF
--- a/src/Drupal/Driver/Fields/Drupal7/DatetimeHandler.php
+++ b/src/Drupal/Driver/Fields/Drupal7/DatetimeHandler.php
@@ -19,7 +19,7 @@ class DatetimeHandler extends AbstractHandler {
   public function expand($values) {
     $return = array();
     if (isset($this->field_info['columns']['value2'])) {
-      foreach ($values as $value) {
+      foreach ((array)$values as $value) {
         $return[$this->language][] = array(
           'value' => $value[0],
           'value2' => $value[1],
@@ -27,7 +27,7 @@ class DatetimeHandler extends AbstractHandler {
       }
     }
     else {
-      foreach ($values as $value) {
+      foreach ((array)$values as $value) {
         $return[$this->language][] = array('value' => $value);
       }
     }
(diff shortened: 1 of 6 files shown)


ademarco:
We should investigate it further since field handlers work on the assumption that values is always an array, and it should since it comes from https://github.com/jhedstrom/drupalextension/blob/master/src/Drupal/DrupalExtension/Context/RawDrupalContext.php#L249-249

Casting to an array is only curing the symptoms, the real problem seems to be somewhere else. What do you think?

dutchiexl:
Strange enough I have different code:

<pre>
public function nodeCreate($node) {
    $this->dispatchHooks('BeforeNodeCreateScope', $node);
    $saved = $this->getDriver()->createNode($node);
    $this->dispatchHooks('AfterNodeCreateScope', $saved);
    $this->nodes[] = $saved;
    return $saved;
  }
</pre>

using this in my composer file:
    "drupal/drupal-extension": "~3.0",

ademarco:
Ah there we go! :) You should test using master branch from both projects.

dutchiexl:
Ok, this may be closed, master works ok

ygerasimov:
How to set master branch properly?

When I have in my behat.yaml file
```
"drupal/drupal-extension": "dev-master",
```

I get error on "composer update"


Problem 1
    - Installation request for drupal/drupal-extension dev-master -> satisfiable by drupal/drupal-extension[dev-master].
    - drupal/drupal-extension dev-master requires drupal/drupal-driver dev-master -> no matching package found.

ygerasimov:
I have updated drupal-driver to master branch manually (downloaded archive from github and replaced it locally), but it didn't help unfortunately.

dutchiexl:
hmmm, this is what I have, and it works for me:
<pre> "drupal/drupal-extension": "dev-master",</pre>
If it doesn't work for you, something must be wrong with the packagist.

jhedstrom:
If you want to use `dev-master` I think you need to set [`minimum-stability`](https://getcomposer.org/doc/04-schema.md#minimum-stability) to `dev`.

dutchiexl:
Yes indeed

dutchiexl:
I added this to my composer:

<pre>
  "minimum-stability": "dev",
  "prefer-stable": true,
</pre>

ygerasimov:
Thanks guys, worked like a charm.

Issue #14, opened 2015-02-25
Defaulthandler problem with array values


dutchiexl:
When I run 
Warning: Invalid argument supplied for foreach() in /var/www/fta/vendor/drupal/drupal-driver/src/Drupal/Driver/Fields/Drupal7/DefaultHandler.php line 21

Because "$values" isn't always an array, it generates the above error.
<pre>
public function expand($values) {
    $return = array();
    foreach ($values as $value) {
      $return[$this->language][] = array('value' => $value);
    }
    return $return;
  }
</pre>

This seems to be a problem for PHP 5.5.9-1ubuntu4.5
but not PHP 5.6.2 (cli)

Casting the variable might solve this:
foreach ((array)$values as $value)

Scenario:
<pre>
    Given I am viewing a "News" content:
      | title      | News feature: News item 1 |
      | body       | This is a news item       |
      | field_date | 2018-02-25 07:45:45       |
      | language   | en                        |
</pre>

PR #22, opened 2015-03-05
Avoid invalid argument DefaultHandler.php non array.

ygerasimov:
Pull request for changes proposed in https://github.com/jhedstrom/DrupalDriver/issues/14.

Fix avoid problem on PHP 5.4.38.


ademarco:
We had a similar pull request not long time ago: https://github.com/jhedstrom/DrupalDriver/pull/19 
Would the solution outlined there work for you as well?

jhedstrom:
I merged #19 as it converted all of the places this was a possibility. Thanks!

PR #23, opened 2015-03-06
Add missing dependency to symfony/dependency-injection

ademarco:
Working with field handlers using DrupalDriver as a standalone library has uncovered a missing dependency to ```symfony/dependency-injection```: this pull request takes care of that by adding it to composer.json.


jhedstrom:
Thanks!

PR #25, opened 2015-03-13
Stay in drupal directory

marcingy:


src/Drupal/Driver/Cores/Drupal8.php
```diff
@@ -33,8 +33,6 @@ public function bootstrap() {
 
     // Initialise an anonymous session. required for the bootstrap.
     \Drupal::service('session_manager')->start();
-
```
jhedstrom: The call to `getcwd()` above can be removed now too (in both 7 and 8 drivers).

jhedstrom:
Thanks!

PR #30, opened 2015-05-08
Use autoload.php in favour of core/vendor/autoload.php

webflo:
Core introduced the autoload.php a while ago. Changed the bootstrap in Drupal8, its follows very close what we have in [index.php](https://github.com/drupal/drupal/blob/8.0.x/index.php#L17).

### DIFF
--- a/src/Drupal/Driver/Cores/Drupal8.php
+++ b/src/Drupal/Driver/Cores/Drupal8.php
@@ -27,7 +27,7 @@ class Drupal8 extends AbstractCore {
 
     // Bootstrap Drupal.
     chdir(DRUPAL_ROOT);
-    $autoloader = DRUPAL_ROOT . '/autoload.php';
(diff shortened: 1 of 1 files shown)


webflo:
The test failures are unrelated to this PR. mockery/mockery requires PHP 5.4 

@jhedstrom Should i change the mockery/mockery constraint to the last 5.3 compatible release?

jhedstrom:
@webflo I removed php 5.3 from travis, but if you want to update the composer.json with the last 5.3-compatible version, that would be awesome!

webflo:
Rebased onto master. I make another PR for 5.3 compatibility. Accordion to packagist is mockery 0.9.4 the last version that supports PHP 5.3.

webflo:
https://packagist.org/packages/mockery/mockery

jhedstrom:
Thanks!

PR #31, opened 2015-05-08
Term Reference Field Handling

kerasai:
The current TaxonomyTermReferenceHandler just plucks the first term found by `taxonomy_get_term_by_name()`. If there are multiple terms with the same name this can cause issues when an entity is assigned a term from the wrong vocabulary.

This patch includes that attempts to determine the vocabulary/bundle for which the field is configured and restricts the `taxonomy_get_term_by_name()` call to said vocabulary.

If unable to field's configured vocabulary, behavior will revert to the current behavior of looking up across all vocabularies.

### DIFF
--- a/src/Drupal/Driver/Fields/Drupal7/TaxonomyTermReferenceHandler.php
+++ b/src/Drupal/Driver/Fields/Drupal7/TaxonomyTermReferenceHandler.php
@@ -30,7 +30,9 @@ class TaxonomyTermReferenceHandler extends AbstractHandler {
   /**
    * Attempt to determine the vocabulary for which the field is configured.
    *
-   * @return string|NULL
+   * @return mixed
+   *   Returns a string containing the vocabulary in which the term must be
+   *   found or NULL if unable to determine.
    */
   protected function getVocab() {
     if (!empty($this->field_info['settings']['allowed_values'][0]['vocabulary'])) {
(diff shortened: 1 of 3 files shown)


ademarco:
Thanks for the initiative! Would it be possible to have a similar approach for Drupal 8 handler?

kerasai:
The code suggests a very similar change to Drupal\Driver\Fields\Drupal8\TaxonomyTermReferenceHandler could be made.

I don't currently have a D8 site in front of me to develop/test against. I'll try to take a look in the near future.

jhedstrom:
Would be great to get this in for the 1.1 release.

kerasai:
I think this is ready to go. Let me know if you have any questions.

D8 implementation took place in the entity reference handler, as the term ref fields are being deprecated.

Also, I was able to demonstrate failing cases for D7 and D8 with the following tests. The changes included in this PR allow these to pass.

**Example Behat Feature:** https://gist.github.com/kerasai/7b5bab855e4e69e3d3b5
**Custom Step for Analyzing Term:** https://gist.github.com/kerasai/ddb24449fe090b7717f3

jhedstrom:
Awesome. Thanks!

pfrenssen:
This broke support for all entities that do not use bundles (for example the `User` entity). There is actually a test for this in DrupalExtension that tests the User entity and this fails now. Unfortunately we weren't able to detect this earlier because of [#193: Tests are failing since Drupal 8 updated Guzzle to 5.3.0](https://github.com/jhedstrom/drupalextension/issues/193).

Will create a followup issue to restore support for entities without bundles.

PR #33, opened 2015-05-21
username_spaces

ryanaslett:
Usernames in stories should be able to have spaces.

### DIFF
--- a/src/Drupal/Driver/DrushDriver.php
+++ b/src/Drupal/Driver/DrushDriver.php
@@ -120,7 +120,7 @@ class DrushDriver extends BaseDriver {
    */
   public function userCreate(\stdClass $user) {
     $arguments = array(
-      $user->name,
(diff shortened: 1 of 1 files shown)


src/Drupal/Driver/DrushDriver.php
```diff
@@ -120,7 +120,7 @@ public function isBootstrapped() {
    */
   public function userCreate(\stdClass $user) {
     $arguments = array(
-      $user->name,
+      sprintf('"%s"', $user->name),
```
jhedstrom: Doesn't this add `"` around the user name (meaning the name would be `"my user name"` rather than `my user name`)?

ryanaslett:
It passes those quotes to the eventual drush command that gets created, so they end up being the quotes on the command line.

jhedstrom:
Oh! I totally missed that this was just in the drush driver. Merged. Thanks!

PR #34, opened 2015-05-30
Skip entity delete for non-existing entities

webflo:
I added some additional checks. The ``Drupal8``has not the same behaviour as ``Druapl6``and ``Drupal7``because ``taxonomy_del_term``, ``taxonomy_term_delete`` and ``entity_delete_multiple`` fail silent on non-existing entities.


jhedstrom:
Thanks!

PR #35, opened 2015-06-03
Drupal7::isField() does not check entity type

pfrenssen:
`Drupal7::isField()` does not check the entity type, it returns `TRUE` if the field exists in any entity.


jhedstrom:
Thanks!

PR #36, opened 2015-06-03
Make field lookup more efficient

pfrenssen:
In `Drupal7::getEntityFieldTypes()` the field lookup is delegated to `isField()` but this is memory inefficient for sites with lots of fields. The necessary data from `field_info_field_map()` is already available in `getEntityFieldTypes()` so we can simply make use of it.

`isField()` does its own call to `field_info_field_map()` and in traditional Drupal 7 fashion this data is represented in an array which is passed through a bunch of functions and methods. Every time an array in passed additional memory will be allocated and the data will be copied in it. We can avoid this and gain a little bit of memory efficiency.


jhedstrom:
Makes sense to me. Thanks!

PR #37, opened 2015-06-05
Fix capitalization of {@inheritdoc}.

pfrenssen:


pfrenssen:
This is integrated in #41 since this was causing code sniffer warnings.

PR #40, opened 2015-06-10
Add support for field columns to the default field handler

pfrenssen:
Currently the default field handler only supports fields that have a single column, and will hard code the column name to 'value'. This is fine for something like a regular text field, but it cannot be used for fields that have multiple columns, or use a different column name.

I propose to allow the use of arrays for the field items in `DefaultHandler::expand()` that are keyed on the column name.

For example this would allow us to support a complex field such as AddressField without having to write a custom handler:

```
$values = [
  0 => [
    'country' => 'BE',
    'locality' => 'Brussel',
    'thoroughfare' => 'Louisalaan 1',
    'postal_code' => '1000',
  ],
];
```

### DIFF
--- a/src/Drupal/Driver/BaseDriver.php
+++ b/src/Drupal/Driver/BaseDriver.php
@@ -78,6 +78,13 @@ abstract class BaseDriver implements DriverInterface {
   /**
    * {@inheritdoc}
    */
+  public function clearStaticCaches() {
+    throw new UnsupportedDriverActionException($this->errorString('clear static caches'), $this);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
   public function createNode($node) {
     throw new UnsupportedDriverActionException($this->errorString('create nodes'), $this);
   }
--- a/src/Drupal/Driver/Cores/CoreInterface.php
+++ b/src/Drupal/Driver/Cores/CoreInterface.php
@@ -191,4 +191,9 @@ interface CoreInterface {
    */
   public function languageDelete(\stdClass $language);
 
+  /**
+   * Clears the static caches.
+   */
+  public function clearStaticCaches();
+
 }
--- a/src/Drupal/Driver/Cores/Drupal6.php
+++ b/src/Drupal/Driver/Cores/Drupal6.php
@@ -491,4 +491,11 @@ class Drupal6 extends AbstractCore {
     throw new \Exception('Deleting languages is not yet implemented for Drupal 6.');
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function clearStaticCaches() {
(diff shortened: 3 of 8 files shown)


pfrenssen:
This is an initial implementation. This still supports the original use case where passing in a scalar value would use the column name 'value' by default.

Personally I would like to get rid of this special case, and simply always require an array to be passed. After all, it is just coincidence that a number of core fields use 'value' as their column name. Many fields use a different name, for example Entity Reference uses 'target_id'. Changing this would be a backwards compatibility break though.

This still needs tests, but I'll start on this after getting feedback.

ademarco:
The original idea was to provide a similar developer experience on D7 as we have now on D8. On D8 one can simply assign:

```php
$node->field_text = 'Text value';
```

Also this works for multiple values:

```php
$node->field_text = array('Text value 1', 'Text value 2');
```

A way to reproduce this behavior on D7 is establishing the following conventions:

##### Single-value assignment

```php
$node->field_text = 'Text value';
```

If a field value is a simple scalar, then it is considered to be a single value field and its value is passed to the driver field handler, which will expand it according to the field type it is used for.

##### Multi-values assignment

```php
$node->field_text = array('Text value 1', 'Text value 2');
```

If a field value is an array with numeric keys then the array's values are processed as the single-value assignment scenario outlined above.

##### Keyed-array single-value assignment

```php
$node->field_link = array('url' => 'http://example.com', 'title' => 'Link title');
```

If a field value is an array with string keys then it is considered as a single-value assignment but having specific field columns.

##### Keyed-array multi-values assignment

```php
$node->field_link = array(
  array('url' => 'http://example.com', 'title' => 'Link title 2'),
  array('url' => 'http://example.com', 'title' => 'Link title 1'),
);
```

If a field value is an array with numeric keys and its values are arrays with string keys then it is considered as a multi-values assignment having each value being treated as the "Keyed-array single-value assignment" case above.


Looking forward to your feedback!

pfrenssen:
@ademarco thanks a lot for the explanation! You're correct, this is simply great DX. So I'll continue on the current path, I'll just have to add a test case then.

pfrenssen:
This will need a reroll now that the coding standards branch is in.

pfrenssen:
Rebased onto the latest master branch, and fixed coding standards warnings :)

pfrenssen:
I'm going to start implementing this in Drupal Extension now. Maybe we can keep this open until that is ready?

jhedstrom:
Sounds good.

pfrenssen:
I will do the work in the [field-columns](https://github.com/pfrenssen/drupalextension/commits/field-columns) branch in my fork. Will make a pull request when it is ready for review.

PR #41, opened 2015-06-11
Check for coding standards on Travis

pfrenssen:
Let's do an automated check for the coding standards on Travis and fix any violations.

### DIFF
--- a/phpcs-ruleset.xml
+++ b/phpcs-ruleset.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0"?>
+<!-- PHP_CodeSniffer standard -->
+<!-- See https://github.com/squizlabs/PHP_CodeSniffer/wiki/Annotated-ruleset.xml -->
+<ruleset name="Invoicing">
+    <description>DrupalDriver coding standard</description>
+
+    <rule ref="./vendor/drupal/coder/coder_sniffer/Drupal">
+    </rule>
+
+    <!-- Core drivers need to set a server's remote address. -->
+    <rule ref="Drupal.Semantics.RemoteAddress.RemoteAddress">
+        <exclude-pattern>src/Drupal/Driver/Cores/*.php</exclude-pattern>
+    </rule>
+
+</ruleset>
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,4 +8,4 @@ install:
 script:
   - phpunit
   - vendor/bin/phpspec run -f pretty --no-interaction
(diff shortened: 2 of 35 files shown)


jhedstrom:
Most of the fails are coming from the docs, which because they are snippets, probably won't ever fully pass. Let's exclude that directory and then fix the remaining fails here.

pfrenssen:
I broke something. Drupal7FieldHandlerTest is failing. This might be because I changed a handful of properties from snake case to camel case.

pfrenssen:
Ready :)

Most of the changes are harmless, but in order to satisfy the coding standards I had to change a number of properties to camelcase:

* `\Drupal\Driver\Fields\Drupal7\AbstractHandler`: `$entityType`, `$fieldName`, `$fieldInfo`.
* `\Drupal\Driver\Fields\Drupal8\AbstractHandler`: `$fieldInfo`.

If this is not acceptable I can revert these changes and add an exception so these checks will be excluded from automated code standards checks.

phpcs-ruleset.xml
```diff
@@ -0,0 +1,15 @@
...
+<ruleset name="Invoicing">
+    <description>DrupalDriver coding standard</description>
+
+    <rule ref="./vendor/drupal/coder/coder_sniffer/Drupal">
+    </rule>
+
+    <!-- Core drivers need to set a server's remote address. -->
+    <rule ref="Drupal.Semantics.RemoteAddress.RemoteAddress">
+        <exclude-pattern>src/Drupal/Driver/Cores/*.php</exclude-pattern>
+    </rule>
```
jhedstrom: Does this rule exclude all core drivers? And if so, why? If they do need to be excluded, can we clarify the reason in the comment (I don't understand the current comment about remote server address).

jhedstrom:
> Most of the changes are harmless, but in order to satisfy the coding standards I had to change a number of properties to camelcase

I think this should be ok. They are already marked as protected, and AFAIK, nobody else is implementing these interfaces yet :)

phpcs-ruleset.xml
pfrenssen: In these files the `$_SERVER['REMOTE_ADDR']` predefined variable is used, which throws the following false positive:

> Use the function ip_address() instead of $_SERVER['REMOTE_ADDR']

By excluding this specific sniff in these files this false positive is suppressed.

jhedstrom:
Awesome, thanks!

PR #43, opened 2015-06-25
Add list_boolean field handler

thePanz:
The list_boolean field type (defined in List core module) is not correctly handled in current implementation: this handler forces the field type to be correctly handled as a "list"


thePanz:
Latest commit failed testing due to a Travis composer timeout.

jhedstrom:
Do you know if this is needed for Drupal 8?

thePanz:
There is no ListTextHandler for Drupal8, thus I think it's not needed for D8 :)

jhedstrom:
Thanks!

PR #44, opened 2015-06-26
Test list boolean field with blank 'On' and 'Off' values.

nlisgo:
I've noticed a bug if I am trying to set the value '0' for a list boolean field when the Off value is blank in the settings.

### DIFF
--- a/src/Drupal/Driver/Fields/Drupal7/ListBooleanHandler.php
+++ b/src/Drupal/Driver/Fields/Drupal7/ListBooleanHandler.php
@@ -9,9 +9,26 @@ namespace Drupal\Driver\Fields\Drupal7;
 
 /**
  * ListBoolean field handler for Drupal 7.
- *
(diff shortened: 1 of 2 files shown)


jhedstrom:
Does this have corresponding equivalents in either Drupal 6 or 8?

nlisgo:
The ListBooleanHandler was added 15 days ago and has just been introduced for Drupal7. I support adding it for D6 and D8 but surely it shouldn't hold up this fix.

https://github.com/jhedstrom/DrupalDriver/pull/43

nlisgo:
If it is not clear, the first failing commit is a test-only commit. I introduced a test to demonstrate the bug and the second commit fixes the bug.

jhedstrom:
Makes sense to me, thanks!

nlisgo:
Thanks for maintaining the project. It's been invaluable to us.

jhedstrom:
Glad you're using it!

PR #46, opened 2015-07-07
Allow for options defined in allowed_values_function

hitfactory:
ListTextHandler updated to allow for the fact that options for a List/Text field can be defined by a function whose name is set in the field's settings array. I can appreciate it's an edge case, but without this I can't create nodes in my tests.


pfrenssen:
Can you provide an example implementation so it is clearer what you are trying to achieve? Why don't you filter the values before passing them in?

If you are using this in combination with the Behat Drupal Extension it is maybe better to fix it there.

In any case this would need to have support for Drupal 8 too and have a test case added.

pfrenssen:
Oh I get it now, this is simply adding support this functionality which is provided by the List module in [list_allowed_values()](https://api.drupal.org/api/drupal/modules!field!modules!list!list.module/function/list_allowed_values/7).

In Drupal 8 this has moved to the Options module, see [options_allowed_values()](https://api.drupal.org/api/drupal/core!modules!options!options.module/function/options_allowed_values/8).

jhedstrom:
Would be great to get this into the 1.1 release. However, a version for 6 and 8 will be needed.

hitfactory:
Thanks for getting back. 

I forgot the function arguments in my PR so need to fix that.

Re: tests, do you have any tips on how best to mock the allowed values function callback?

jhedstrom:
If the callback function takes any callable (eg `MyTestClass::foo`), then perhaps that could be used for the test?

jhedstrom:
Since this has immediate benefit to 7, and doesn't impact the public API, I think it is good to go in without 6 or 8 support (those can be added if/when needed). What do others think?

jhedstrom:
Since nobody seems to object, I'm merging this for the reasons mentioned above. Follow-ups for 6, and/or tests are welcome.

PR #49, opened 2015-07-27
Add support for Image fields in Drupal 7

acbramley:
This adds an ImageHandler class to parse an array of values and add the required attributes for an image field. Have tested this locally using beforeNodeCreation hook like so

````
  /**
   * Hook into node creation to test `@beforeNodeCreate`
   *
   * @beforeNodeCreate
   */
  public function alterNodeParameters(BeforeNodeCreateScope $scope) {
    $node = $scope->getEntity();
    // Parse the node image.
    if (isset($node->main_image)) {
      $value = $node->main_image;
      unset($node->main_image);

      $path = $this->getMinkParameter('files_path') . $value;
      $file_contents = file_get_contents($path);

      $dirname = "public://behat_test_images";
      if (file_prepare_directory($dirname, FILE_CREATE_DIRECTORY)) {
        $file = file_save_data(
          $file_contents,
          $dirname . '/' . $value,
          FILE_EXISTS_REPLACE
        );
        $file_properties = array(
          'filename' => $file->filename,
          'uri' => $file->uri,
          'fid' => $file->fid,
          'display' => 1
        );
        $node->field_image = implode(' - ', $file_properties);
      }
    }
  }
````

I feel like the simplicity of my approach might reflect my little understanding about this library so if I've gone about it completely wrong please let me know. This worked with the drupaldriver as well when adding a hyphen separated string into a "Given :type nodes:" step.

Cheers

### DIFF
--- a/src/Drupal/Driver/Fields/Drupal7/ImageHandler.php
+++ b/src/Drupal/Driver/Fields/Drupal7/ImageHandler.php
@@ -1,0 +1,31 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\Driver\Fields\Drupal7\ImageHandler
+ */
+
+namespace Drupal\Driver\Fields\Drupal7;
+
+/**
+ * Class ImageHandler
+ * @package Drupal\Driver\Fields\Drupal7
+ */
+class ImageHandler extends AbstractHandler {
+
+  /**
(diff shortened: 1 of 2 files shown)


acbramley:
Closing as was incorrectly using master

PR #52, opened 2015-08-14
Fixes issue where existing terms with the same name were returns from termCreate()

### LINKED ISSUE #50: Drupal7 termCreate returns existing terms of the same name
I've just tracked down an issue we've been having for a while, some of our terms were being periodically deleted - this turns out to be an issue with how the Drupal7 Driver creates terms and what it subsequently returns.

The issue is that after doing the taxonomy_term_save() it unnecessarily reloads terms by using taxonomy_get_term_by_name and uses a reset() to return the first term from the list.

This is unnecessary because $term is an object and will have the resulting saved term after taxonomy_term_save() runs. Even though the comments state "there should only be one matching term in a testing context" that can never be safely true, what about free tagging vocabularies which contain user created terms?

I propose we remove the reloading altogether and return $term straight after the save.

I will have a pull request up for this soon.

acbramley:
Fixes #50

### DIFF
--- a/src/Drupal/Driver/Cores/Drupal7.php
+++ b/src/Drupal/Driver/Cores/Drupal7.php
@@ -324,13 +324,7 @@ class Drupal7 extends AbstractCore {
 
     \taxonomy_term_save($term);
 
-    // Loading a term by name returns an array of term objects, but there should
-    // only be one matching term in a testing context, so take the first match
-    // by reset()'ing $matches.
-    $matches = \taxonomy_get_term_by_name($term->name);
-    $saved_term = reset($matches);
-
-    return $saved_term;
+    return $term;
   }
 
   /**


src/Drupal/Driver/Cores/Drupal7.php
```diff
@@ -324,13 +324,7 @@ public function termCreate(\stdClass $term) {
...
     \taxonomy_term_save($term);
 
-    // Loading a term by name returns an array of term objects, but there should
-    // only be one matching term in a testing context, so take the first match
-    // by reset()'ing $matches.
-    $matches = \taxonomy_get_term_by_name($term->name);
-    $saved_term = reset($matches);
-
-    return $saved_term;
+    return $term;
```
jhedstrom: Does this get the term ID from the newly created term?

acbramley: @jhedstrom yeah, since $term is an object it gets the tid after save. The same happens with menu_link_save(), node_save(), etc. I've tested it with a fairly complex test suite and it worked well.

You can even see in the taxonomy_term_save function it uses the tid from the $term object to insert into the hierarchy table (https://api.drupal.org/api/drupal/modules!taxonomy!taxonomy.module/function/taxonomy_term_save/7)

jhedstrom:
Thanks for verifying, and for the fix!

acbramley:
Thanks a lot @jhedstrom !

PR #53, opened 2015-08-24
Field support is broken for entities without bundles on Drupal 8

pfrenssen:
In #31 support for taxonomy term reference fields was added, but as an unintentional side effect the support for bundleless entities (such as the `User` entity) was broken.

There is test coverage for this use case in the Behat Drupal Extension and this fails now. Unfortunately we weren't able to detect this earlier because of an issue in BDE preventing us from successfully running the test suite on D8 (see [#193: Tests are failing since Drupal 8 updated Guzzle to 5.3.0](https://github.com/jhedstrom/drupalextension/issues/193)).


jhedstrom:
Thanks!

PR #56, opened 2015-09-01
Enhance the Drush driver to allow creation of nodes and taxonomy terms.

greg-1-anderson:
This is still a prototype, submitted with the goal of collecting feedback on the concept and direction. It is super useful to be able to create nodes and taxonomy terms with the Drush Driver, e.g. when writing Behat tests on Circle or Travis, and testing on a remote Drupal site.

This PR adds additional 'drush' calls to the Drush driver to allow for the creation of nodes and taxonomy terms.  These capabilities are not provided by default in Drush core; therefore, to make this work, a new Drush commandfile, behat.d7.drush.inc, must be installed in the target Drupal site.  This commandfile handles all node and term create and delete requests.

The current implementation of behat.d7.drush.inc uses some classed from the DrupalDriver.  This only works if you have set up your Drupal site such that it is managed by Composer, and both Drush and Behat (and, of course, the Behat DrupalDriver) are required from your composer.json.  If you do this, then Drush will load the autoload file that contains the Behat DrupalDriver classes.  Other parts of the code are copied in, rather than used directly.  I think that it would be necessary to set up a Symfony container in order to use the other classes here, although clearly I did not try this.  drush-ops/drush#1377 would be useful here, though.

There is also a behat.d8.drush.inc file that is mentioned in this PR, but not provided; this still needs to be implemented.

You will also notice that there is minimal (no) error checking inside the new methods in the Drush driver.  Currently, if the behat Drush extension is not loaded in the target site, this causes an exception to be thrown, which is already reported.  It would probably be possible to catch this exception, and replace the reported error message with advice / instructions on how to install the Behat Drush extension.


greg-1-anderson:
The Travis failures are just lint warnings; I fixed these up in the second commit, but Travis re-tested the first commit again instead, for some reason.

jhedstrom:
This is great @greg-1-anderson! It will make a lot of folks who are unable to run Behat on the same server as their application very happy. I wonder if there is any way to add tests for this to `.travis.yml` (or alternatively, add phpunit tests for the drush commands, or both).

greg-1-anderson:
Hm, I don't see an easy way to add very much value here with phpunit tests. There are no existing tests for create node & c., and this would be a little hard to do with a unit test.

I think the biggest value would be to use functional testing here. In travis.yml, download Drush and Behat, use drush to install a Drupal site or two, point the Drupal / Drush driver to the branch here being tested, and then run some functional tests.  The functional tests would be testing Drupal behavior, but the test title / description would name the parts of the driver being exercised.  I'd sprint on this at BadCamp with you, if you're interested.

When are you planning on releasing 1.2.0?  I'll try to make a behat.d8.drush.inc pretty soon.

greg-1-anderson:
Of course, it would be possible to make much more direct tests for the Drush commands, but again we'd need to use Drush + Drupal to do it.

greg-1-anderson:
I think that I am going to make the behat.*.drush.inc a separate GitHub repository, so that we can put it in Composer, and have it just drop into the right location when using Composer to manage your Drupal sites.  We can also then simply 'require' it from the composer.json of DrupalDriver.

Does that sound like a good idea to you?

jhedstrom:
Testing directly via functional tests sounds good. I've yet to add those here, but that's what the [Drupal Extension](https://github.com/jhedstrom/drupalextension) does for its tests.

Regarding `composer.json`, this sounds great. However, wouldn't this would be a documentation issue, since adding it to the Drupal Driver's `composer.json` would simply place those...somewhere, I'm not actually sure :)

greg-1-anderson:
Yes, true, in many instances, the files would just get buried in vendor (maybe vendor/drush/BehatDrushCommand).  However, if the top-level composer.json was a Drupal site, and it had the appropriate custom installer directives in it, then the files would end up at __ROOT__/drush/BehatDrushCommand, or __ROOT__/sites/all/drush/BehatDrushCommand, which would be super-easy-exactly-right for those using Composer this way.

For the 'buried' case, we could always document where to look for the files.  If the DrushDriver got an error, it could even calculate where the Behat Drush Command was found, and where it should go, and could print out the appropriate cp -R or rsync command for the user.

greg-1-anderson:
Started an independent project for the Drush commandfiles: https://github.com/pantheon-systems/behat-drush-endpoint

greg-1-anderson:
Updated this PR to catch up with the factored-out Drush commandfiles.  Still pointing at pantheon-systems, but planning to relocate; will update in a future commit.

greg-1-anderson:
Relocated behat-drush-endpoint to https://github.com/drush-ops/behat-drush-endpoint.

There are instructions in that project how to set things up to use this feature; it might be easier / better to add functional tests with at least a dev release of this, though.

composer.json
```diff
@@ -19,7 +19,8 @@
     "drupal/coder": "~8.2.0",
     "phpspec/phpspec": "~2.0",
     "phpunit/phpunit": "~4.0",
-    "mockery/mockery": "0.9.4"
+    "mockery/mockery": "0.9.4",
+    "drush-ops/behat-drush-endpoint": "*"
```
jhedstrom: Since this is now documented as a separate project, with separate installation instructions, is there any point in including it here?

greg-1-anderson: We could remove it; the advantage of removing it is that not all users need it.  The advantage of including it here is that the separate installation instructions do not need to be followed if the user is managing their site with Composer -- the Drush Driver will just start working as of the next `composer update`.  The separate installation instructions will need to be followed by users who are not managing their Drupal sites with Composer, though.

It's a small extension; I'd recommend keeping it in the dependencies here.  It's completely up to you, though.

jhedstrom: This makes sense.

jhedstrom:
I think this PR is good to go. Once this is in, I think the BDE tests/example docs could use updating.

jhedstrom:
Thanks!

greg-1-anderson:
Awesome! Thanks!

PR #58, opened 2015-09-15
D7 - Update the roles just added.

clemens-tolboom:
The current user is used to validate field permissions.

The current $user has no roles field so when used fails in user.module

drupal7_php

### DIFF
--- a/src/Drupal/Driver/Cores/Drupal7.php
+++ b/src/Drupal/Driver/Cores/Drupal7.php
@@ -134,6 +134,9 @@ class Drupal7 extends AbstractCore {
     }
 
     user_multiple_role_edit(array($user->uid), 'add_role', $role->rid);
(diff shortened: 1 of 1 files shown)


clemens-tolboom:
> Notice: Undefined property: stdClass::$roles in /Users/clemens/Sites/ctcn/dev/www/modules/user/user.module line 821

user_module-roles

which uses the global $user to check for field permissions.

Snippet below uses og_group field triggering entity_reference

```php
    // Creating the node will automatically redirect the browser to view it.
    $fields = new TableNode([
      ['title', $title . 'user-' . $user->uid],
      [$group_field, $group->title],
      ['uid', $this->drupalContext->user->uid]
    ]);
    $this->drupalContext->assertViewingNode($type, $fields);
```

jhedstrom:
Thanks!

PR #61, opened 2015-10-08
Make Drupal 7 and 6 drivers return the rid from roleCreate()

bircher:
The interface defines:
```php
/**
   * Create a role
   *
   * @param array $permissions
   *   An array of permissions to create the role with.
   *
   * @return integer
   *   The created role name.
   */
  public function roleCreate(array $permissions);
```

this contract is broken by the Drupal 7 driver that returns the role name.
This then in turn breaks the DrupalExtension behat step:
```php
/**
   * @Given I am logged in as a user with the :permissions permission(s)
   */
  public function assertLoggedInWithPermissions($permissions){}
```


bircher:
actually the interface is also confused!
It says integer and then the description is role name...

The Drupal 8 does return ```return $role->id();```
So it is the string then!

PR #65, opened 2015-10-08
Implement clear static caches on CoreInterface

dawehner:


src/Drupal/Driver/DrupalDriver.php
```diff
@@ -140,6 +140,14 @@ public function clearCache($type = NULL) {
   /**
    * {@inheritdoc}
    */
+  public function clearStaticCaches() {
```
jhedstrom: This class already has this method down near the bottom. The issue is that the method was added here without being added to the interface.

jhedstrom:
I removed the duplicated methods in #66. Marking this as closed.

PR #67, opened 2015-11-05
Add an 'isField' method to the Drush Driver.

greg-1-anderson:
Fixes failures related to tests that create nodes & c. with fields.


jhedstrom:
Thanks!

PR #71, opened 2015-12-04
File field handler usage

becw:
This pull request is half question, half suggestion: how is the image field handler intended to be used? When I try to use it, I have to know an existing `fid`, which is generally not possible from the context where I'm using it (in a Behat test).

For my use cases, it would make sense for it to behave more similarly to the `EntityreferenceHandler`, using filename values to match a Drupal-managed file, if available. This will require some additional code in the cores, and probably also in `jhedstrom/drupalextension`.

This PR adds `\Drupal\Driver\Fields\Drupal7\FileHandler` with filename-based field values, and tweaked `\Drupal\Driver\Fields\Drupal7\ImageHandler` to extend on that. This does break with the current behavior.

In order to be complete, this PR would need the following (but I didn't want to proceed if there's a better way to use the `ImageHandler`):

- [ ] Make sure that the filename lookup works for D8 as well as D7
- [ ] Add file creation and deletion to the D7 and D8 cores
- [ ] Add file creation and cleanup to the drupal-extension

Does this approach make sense?

### DIFF
--- a/src/Drupal/Driver/Fields/Drupal7/FileHandler.php
+++ b/src/Drupal/Driver/Fields/Drupal7/FileHandler.php
@@ -1,0 +1,49 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\Driver\Fields\Drupal7\FileHandler.
+ */
+
+namespace Drupal\Driver\Fields\Drupal7;
+
+/**
+ * File field handler for Drupal 7.
+ */
+class FileHandler extends AbstractHandler {
+
+  /**
+   * {@inheritdoc}
+   *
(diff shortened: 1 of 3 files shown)


jhedstrom:
I think this approach makes sense. Being able to use the filename makes much more sense that having to know the `fid`. It is similar to how some of the term and user methods work (eg, term name instead of `tid`).

acbramley:
Looks like a good enhancement, I can't remember what my intentions were around the original ImageHandler code, but this is an awesome step forward!

jhedstrom:
Thanks!

PR #75, opened 2015-12-17
Fix d8 fields

greg-1-anderson:
The changes in this PR enable:

- Setting the author of a created node
- Setting the body text of a created node


greg-1-anderson:
This has been performing well in ad-hoc testing. n.b. the uid change was also made in the behat-drush-endpoint, already committed and tagged (0.0.5).  If you merge this in, it is likely that the drupalextension drush tests will get farther.  It is a little inconvenient to run those locally, though, so I have not tried them directly yet.

jhedstrom:
Thanks!

PR #76, opened 2016-01-15
Adds support for config changes via driver

acbramley:
A follow up drupalextension PR will be added for implementing step definitions :)

### DIFF
--- a/src/Drupal/Driver/Cores/CoreInterface.php
+++ b/src/Drupal/Driver/Cores/CoreInterface.php
@@ -196,4 +196,29 @@ interface CoreInterface {
    */
   public function clearStaticCaches();
 
+  /**
+   * Returns a configuration item.
+   *
+   * @param string $name
+   *   The name of the configuration object to retrieve.
+   * @param string $key
+   *   A string that maps to a key within the configuration data.
+   *
+   * @return mixed
+   *   The data that was requested.
+   */
+  public function configGet($name, $key = '');
+
+  /**
+   * Sets a value in a configuration object.
+   *
+   * @param string $name
+   *   The name of the configuration object.
+   * @param string $key
+   *   Identifier to store value in configuration.
+   * @param mixed $value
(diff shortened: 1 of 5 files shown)


acbramley:
Only reason I haven't done D6 or D7 is because the variable_get and set functions don't really match up with how D8 passes parameters around so more thought needs to go into that.

pfrenssen:
Indeed for Drupal 6 and Drupal 7 the direct analog is getting and setting variables. Since we are providing an abstraction layer we should probably figure the compatibility between D6-D7-D8 before merging this. If we merge this as is we won't be able to change it later without causing a BC break.

The main difference is that a config item is specified with a combination of two parameters in D8: `$name` and `$key`, in D6 and D7 it is only `$name`.

Now since config is chainable on D8 we might get away with using only `$name` in `CoreInterface::configGet()` and `::configSet()`. For D8 we can split off the key from the name at the last period. E.g. `views.view.who_s_new` would get the name `view.view` and the key `who_s_new`.

For me personally I like the fact that it mentions config, and that we would use this naming system also for D6 and D7. I always found "variable" to be a very poor name for what is essentially config.

acbramley:
Yeah, that's exactly what I had come to as well.

The problem with splitting off the last period is with nested config. For example you can have:

my_module.settings.yml:
````
title: Things
stuff:
  foo: bar
````

To get the value of foo with drush you have to do a:
````drush config-get my_module.settings stuff.foo````

If we put that all into 1 string, the "name" would end up my_module.settings.stuff and the "key" would be foo, which doesn't work.

I guess this could be a non-issue with the way I've allowed JSON strings in complex config.

jhedstrom:
This is great. Thanks! I think we can address the 6/7 stuff in a follow-up if and when somebody comes along that needs the functionality there.

acbramley:
@jhedstrom cheers! The DrupalExtension support is here https://github.com/jhedstrom/drupalextension/pull/244

PR #77, opened 2016-01-29
Fix drush nocolor option

lmakarov:
This PR addresses several issues:

1. The hardcoded `--nocolor=` option does not work. It has to be passed either as `--nocolor` or `--nocolor=1`
2. It is not possible to fix/override ^^ via `global_options` in `behat.yml`, since hardcoded options are injected after the global ones.


jhedstrom:
Thanks!

PR #79, opened 2016-03-17
improved logic to avoid using the optional select list label in the  …

eknapier:
…reference array

### DIFF
--- a/src/Drupal/Driver/Fields/Drupal7/ListTextHandler.php
+++ b/src/Drupal/Driver/Fields/Drupal7/ListTextHandler.php
@@ -17,13 +17,23 @@ class ListTextHandler extends AbstractHandler {
    */
   public function expand($values) {
     $return = array();
-    $options = $this->fieldInfo['settings']['allowed_values'];
     if (!empty($this->fieldInfo['settings']['allowed_values_function'])) {
       $cacheable = TRUE;
       $callback = $this->fieldInfo['settings']['allowed_values_function'];
-      $options = call_user_func($callback, $this->fieldInfo, $this, $this->entityType, $this->entity, $cacheable);
+      $allowed_values = call_user_func($callback, $this->fieldInfo, $this, $this->entityType, $this->entity, $cacheable);
     }
-    $allowed_values = array_flip($options);
+    else {
+      $allowed_values = array();
+      $options = array_flip($this->fieldInfo['settings']['allowed_values']);
+      foreach ($values as $value) {
+        if (array_key_exists($value, $options)) {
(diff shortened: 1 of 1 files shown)


eknapier:
This pull request addresses an issue I encountered when creating nodes with required fields whose values are defined in the Allowed Values List section of field settings. The existing technique in DrupalDriver ListTextHandler.php of using array_flip() on the reference array requires me to use the label as value in order to set the intended value on the field. This pull request allows me to create the node using the real field value instead of the label.

jhedstrom:
Are you using this with the Drupal Extension? If so, would it be possible to demonstrate steps to reproduce (and maybe even add a test to that project that leverages this fix)?

eknapier:
I am using the Drupal Extension, but I am also developing some classes to interact directly with the DrupalDriver. The current issue I'm addressing does not involve the Drupal Extension, I will explain the issue below.

Following is sample markup for the Allowed Values List from my field settings for a field "field_sample_values":
`value1|This is an optional label for value1 that could change or be removed.
value2|This is an optional label for value2 that could change or be removed.
value3|This is an optional label for value3 that could change or be removed.`

_note the help text for this section:_
`The possible values this field can contain. Enter one value per line, in the format key|label.
The key is the stored value. The label will be used in displayed values and edit forms.
The label is optional: if a line contains a single string, it will be used as key and label.`

In the processing of these values for nodes, the DrupalDriver is using array_flip() which inverts the key/value relationship of the original array and forces code that creates new nodes with values on this field to reference the label instead of the value. For example:

`$node = new \stdClass();
$node->title = 'Test Sample Values';
$node->type = "article";
$node->body['und'] = 'lorem ipsum';
$node->status = 1;
$node->field_sample_values['und'] = 'This is an optional label for value1 that could change or be removed.';
$node = $this->getDriver()->createNode($node);`

This code above, when executed,  sets 'value1' for 'field_sample_values'.

With my Pull Request in place I can change this code how I want it to reference the real field value:
`$node->field_sample_values['und'] = 'value1';`

My code simply removes the use of the label as key, replacing the reference array currently in use with an array having matching keys & values. This change allows the code to continue returning the same correct value while allowing nodeCreate() to set the expected value for this field from other code.

jhedstrom:
@eknapier I can't quite tell if this change maintains backward compatibility--would code out there that is using the *label* still function as expected?

eknapier:
@jhedstrom My original Pull Request did not maintain backward compatibility but I have changed it so that now it does. Thanks for the feedback. I have tested locally and with the latest Pull Request in place I can set the field by either label or value.

jhedstrom:
Thanks!

PR #80, opened 2016-04-08
Update phpunit example with setUpBeforeClass

bleeDev:


jhedstrom:
Test failures are due to an apparent coding standard change: https://www.drupal.org/node/2665744

Thanks!

PR #83, opened 2016-04-09
Throw an exception if the node type is missing or does not exist

mlncn:
Used "content" and "content type" in the exception messages; i don't think developers will be confused and potential non-developers getting the exception because of a Behat test referring to a non-existent content type will immediately know what to fix.

Related to https://github.com/jhedstrom/drupalextension/issues/265


jhedstrom:
Thanks!

PR #85, opened 2016-05-30
Adding node and user alter functionality

aronbeal:
This is a continuation of the thread [here](https://github.com/jhedstrom/drupalextension/issues/276).  I had thought that I didn't need alteration functionality, but I've expanded the extension capabilities to the point where I needed to have this thing in the core driver, and I felt it might be a useful general addition.

I'm currently using my own fork of DrupalDriver within DrupalExtension that contains this code. I think it's a useful general addition, so I'm submitting the pull request.  The alter functions expect fully loaded drupal objects for the first argument.

This is currently a D7 only implementation.  I'm game to implement in D8 as well eventually, when we aren't so under the gun at my shop. 

One area I need to highlight is the modification to userCreate to deal with the handling for user roles.  I'm not sure that exploding the string there is the best solution to the problem - I welcome input.

### DIFF
--- a/src/Drupal/Driver/BaseDriver.php
+++ b/src/Drupal/Driver/BaseDriver.php
@@ -45,6 +45,20 @@ abstract class BaseDriver implements DriverInterface {
   /**
    * {@inheritdoc}
    */
+  public function nodeDeleteMultiple(array $nids){
+    throw new UnsupportedDriverActionException($this->errorString('node delete multiple'), $this);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function userDeleteMultiple(array $uids){
+    throw new UnsupportedDriverActionException($this->errorString('user delete multiple'), $this);
+  }
+
+  /**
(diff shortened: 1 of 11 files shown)


aronbeal:
Note: in addition to the above, I've also added support for deleting multiple nodes and users at once to the interfaces, base classes (that throw 'non-implement' exceptions), and to D7.  I don't (yet) have the ability to add and test the functionality for a D8 instance.

jhedstrom:
Are all the entity load methods needed for the alter functionality, or did they sneak in from another branch/PR?

AlexSkrypnyk:
Hey @aronbeal - thank you for this contribution! Drupal 7 has reached end of life and we will be dropping D6/D7 support in DrupalDriver 3.x (see #312). Since this PR primarily targets D6/D7 drivers and handlers, I'm closing it. If this functionality is still needed for Drupal 10/11, feel free to open a new issue. Appreciate the work!

PR #86, opened 2016-05-31
Added Drupal8 Image handler for image field types. 

mroest:
It supports both remote as local images.
Created this to add support in my behat tests

### DIFF
--- a/src/Drupal/Driver/Fields/Drupal8/ImageHandler.php
+++ b/src/Drupal/Driver/Fields/Drupal8/ImageHandler.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace Drupal\Driver\Fields\Drupal8;
+
+/**
+ * Image field handler for Drupal 7.
+ */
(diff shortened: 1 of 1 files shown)


src/Drupal/Driver/Fields/Drupal8/ImageHandler.php
```diff
@@ -0,0 +1,38 @@
...
+
+    if (FALSE === $file) {
+      throw new \Exception("Error saving file");
+    }
+
+    $file->save();
+
+    $return = array(
+      'target_id' => $file->id(),
+      'alt' => 'Behat test image',
```
jhedstrom: Is there a way to make these the default value (`alt` and `title`) in case somebody wanted to test with actual values?

jhedstrom:
I went ahead and merged this. We can add support for custom `alt` and `title` attributes in the future if anybody needs that level of customization.

jhedstrom:
Thanks!

PR #87, opened 2016-06-03
Update Drupal7.php

aronbeal:
This came about as I was testing the driver independently of the Travis CI environment, and couldn't figure out why I kept getting such strange errors from the languageCreate D7 function with my basic site install.  Come to find out that without the locale module, the languages table doesn't even exist, and boostrapping locale.inc won't make the 'locale_language_list' function available without that module being enabled.  

This just adds a check for the 'locale' module before doing any language work.  Probably very edge case, but why not.


jhedstrom:
Thanks!

PR #89, opened 2016-06-07
Adds embridge asset item field handler.

acbramley:
Support for the embridge module (drupal.org/project/embridge) for Druapl 8.

Works the same as the EntityReferenceHandler.

Requires custom step definitions to load the assets into drupal + emdb, for example:

```
/**
   * Creates embridge assets.
   *
   * Table in the form:
   * | filename |
   * | cat.png  |
   * | ...      |
   *
   * @Given embridge assets:
   */
  public function createAssets(TableNode $asset_table) {
    $mime_type_guesser = \Drupal::service('file.mime_type.guesser');
    $entity_type_manager = \Drupal::service('entity_type.manager');
    $embridge_client = \Drupal::service('embridge.client');
    $catalogs = EmbridgeCatalog::loadMultiple();
    $catalog = array_pop($catalogs);

    $asset_ids = [];
    foreach ($asset_table->getHash() as $asset_hash) {
      $filename = $asset_hash['filename'];
      $file_path = DRUPAL_ROOT . '/../tests/behat/test_files/' . $filename;
      if (!file_exists($file_path)) {
        throw new \Exception(sprintf('File %s does not exist in the images directory.', $filename));
      }

      // Begin building file entity.
      $values = [
        'uid' => 1,
        'filename' => $filename,
        'source_path' => $file_path,
        'filesize' => filesize($file_path),
        'catalog_id' => $catalog->id(),
      ];
      $values['filemime'] = $mime_type_guesser->guess($values['filename']);

      // Create the local Embridge asset entity.
      /** @var \Drupal\Core\Entity\EntityStorageInterface $asset_storage */
      $asset_storage = $entity_type_manager->getStorage('embridge_asset_entity');
      /** @var \Drupal\embridge\EmbridgeAssetEntityInterface $asset */
      $asset = $asset_storage->create($values);
      // Try to upload the file.
      try {
        $embridge_client->upload($asset);
        $asset->save();
      }
      catch (\Exception $e) {
        $error_message = $e->getMessage();
        throw new \Exception(sprintf("File upload of %s failed with: %s", $filename, $error_message));
      }
      $asset_ids[] = $asset->id();
      $this->embridge_assets[] = $asset;
    }

    return $asset_ids;
  }

  /**
   * Remove any created embridge assets.
   *
   * @AfterScenario
   */
  public function cleanEmbridgeAssets() {
    foreach ($this->embridge_assets as $asset) {
      $asset->delete();
    }
    $this->embridge_assets = [];
  }
```

### DIFF
--- a/src/Drupal/Driver/Fields/Drupal8/EmbridgeAssetItemHandler.php
+++ b/src/Drupal/Driver/Fields/Drupal8/EmbridgeAssetItemHandler.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace Drupal\Driver\Fields\Drupal8;
+
+/**
+ * Embridge Asset Item field handler for Drupal 8.
+ */
+class EmbridgeAssetItemHandler extends EntityReferenceHandler {
+
+}


jhedstrom:
I'm not sure I want to go down the road of adding support for contrib into this. I wonder if there's a way in the next major version to re-architect and move towards a plugin system to make adding this sort of thing easier. Thoughts?

jhedstrom:
On the other hand, this is incredibly simple in its current implementation...

acbramley:
@jhedstrom yeah I half expected that would be the case but I couldn't think of a way to nicely add this to my codebase. No worries if you want to decline it, I can still use the PR as a patch.

jhedstrom:
Let's leave this open so we can figure out a more generalizable way of doing things like this.

jhedstrom:
Nearly 2 years later and no easier way to do this. Since it was so incredibly simple I've merged this PR. Thanks!

PR #91, opened 2016-06-09
Remove cruft (that can happen with remote aliases) from drush output …

dsnopek:
… before decoding JSON.

### DIFF
--- a/src/Drupal/Driver/DrushDriver.php
+++ b/src/Drupal/Driver/DrushDriver.php
@@ -191,7 +191,8 @@ class DrushDriver extends BaseDriver {
   /**
    * Decodes JSON output returned by Drush.
    *
-   * It will clean up any junk that may have appeared before the JSON.
+   * It will clean up any junk that may have appeared before or after the
+   * JSON object. This can happen with remote Drush aliases.
    *
    * @param string $output
    *   The output from Drush.
@@ -201,6 +202,8 @@ class DrushDriver extends BaseDriver {
   protected function decodeJsonOutput($output) {
     // Remove anything before the first '{'.
     $output = preg_replace('/^[^\{]*/', '', $output);
+    // Remove anything after the last '{'.
+    $output = preg_replace('/[^\}]*$/s', '', $output);
     return json_decode($output);
   }
 
--- a/src/Drupal/Driver/DrushDriver.php
+++ b/src/Drupal/Driver/DrushDriver.php
@@ -189,11 +189,27 @@ class DrushDriver extends BaseDriver {
   }
 
   /**
+   * Decodes JSON output returned by Drush.
+   *
+   * It will clean up any junk that may have appeared before the JSON.
+   *
+   * @param string $output
+   *   The output from Drush.
+   * @return object
+   *   The decoded JSON object.
+   */
+  protected function decodeJsonOutput($output) {
+    // Remove anything before the first '{'.
+    $output = preg_replace('/^[^\{]*/', '', $output);
+    return json_decode($output);
+  }
+
+  /**
    * {@inheritdoc}
    */
   public function createNode($node) {
     $result = $this->drush('behat', array('create-node', escapeshellarg(json_encode($node))), array());
-    return json_decode($result);
+    return $this->decodeJsonOutput($result);
   }
 
   /**
@@ -208,7 +224,7 @@ class DrushDriver extends BaseDriver {
    */
   public function createTerm(\stdClass $term) {
     $result = $this->drush('behat', array('create-term', escapeshellarg(json_encode($term))), array());
-    return json_decode($result);
+    return $this->decodeJsonOutput($result);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/Drupal/Driver/DrushDriver.php
```diff
@@ -189,11 +189,27 @@ public function clearStaticCaches() {
...
+   * It will clean up any junk that may have appeared before the JSON.
+   *
+   * @param string $output
+   *   The output from Drush.
+   * @return object
+   *   The decoded JSON object.
+   */
+  protected function decodeJsonOutput($output) {
+    // Remove anything before the first '{'.
+    $output = preg_replace('/^[^\{]*/', '', $output);
```
jhedstrom: Should we remove anything after the last `}` too?

dsnopek: Ack, yeah, probably... That's a harder regex to write, though - I'm not sure how to make sure it goes from the last `}` and not the first `}`. I suppose we could reverse the order of the string, check from the beginning, and then reverse it back to the original order? Anyone know regexes better than me know a way to do it?

dsnopek: I don't think the 's' modifier is needed because we aren't using `.` in the regex. The docs say:

> A negative class such as [^a] always matches a newline character, independent of the setting of this modifier.

However, that still looks like to me like it'll match up until the first `}` rather than the last `}`, but I'd be happy to be wrong. :-)

dsnopek: I forgot to add, that I'll test this later (probably tomorrow) when I have time to come back to this :-)

greg-1-anderson: You are right that we do not need the 's' modifier here.

`[^\}]` will match anything except `}`, so it will not match after the first `}`.  The `$` ensures that the matching pattern must be at the end of the buffer.

greg-1-anderson: Here is a crazy regex that strips in one pass (tested):

`$output = preg_replace('/(^[^\{]*|[^\}]*$)/', '', $output);`

Doing it in two lines might be a lot clearer, though.

dsnopek: @greg-1-anderson Your regex wizardry worked! On thinking through it again, it does finally make sense to me. With [your first regex](https://github.com/jhedstrom/DrupalDriver/pull/91/files/bc98ee0ef2795df76b28e6189361c0461e4c97c8#r66506455), the engine would probably start capturing immediately, and each time it met a `}` it would clear what it had previously captured, and start capturing again, until ultimately it met the end of the buffer and the last bit that it captured would be the match. Or some variation on that. Regexes are hard. :-)

I've updated the PR. I chose to do it in two separate lines for clarity, but I'd be happy to switch it to the one line version if you guys preferred! Thanks again!

jhedstrom:
Thanks!

dsnopek:
Woohoo! Thanks for the merge :-)

jhedstrom:
Hmm, this appears to have broken field support for the behat drush endpoint. Note the body field no longer is created properly.

https://travis-ci.org/jhedstrom/drupalextension/jobs/138205688#L708

jhedstrom:
I had a bit of time to do some debugging, and *before* this pr, `$node->body = [0 => 'a placeholder']` comes into drush as an array, wheras after this pr, it comes in as just `$node->body = 'a placeholder'`. I haven't had a chance to debug further. cc @greg-1-anderson @dsnopek

dsnopek:
Interesting! I can't think of how this could have that effect, but I'll guess we'll need to look at the JSON text before it's decoded to see what the difference is...

dsnopek:
I found the problem. The 'is-field' command is return a JSON boolean (ie. `true` or `false`) not a JSON object and the `decodeJsonOutput()` method was assuming a JSON object, and anything that wasn't a JSON object was cruft that was removed. :-)

Here's a new PR:

https://github.com/jhedstrom/DrupalDriver/pull/92

It makes it clear that internal method is just for JSON objects by renaming it to `decodeJsonObject()` and it uses a little cleverness to read the boolean in a way that works regardless of weather there's cruft or not (which there will be with a remote alias).

I think it would actually be cleaner if the drush-behat-extension *always* returned a JSON object. We can't really deal generically with any type of JSON output without getting Drush to pipe stuff into the right streams with remote aliases (ie. the cruft would go to stderr and we'd just have JSON in stdout).

greg-1-anderson:
We have wanted to clean up the mishandled pipes in Drush for a long time, but no one has done it yet. Correcting the handling of stderr would have to be reserved for Drush 9. For earlier versions of Drush, we could add another option that would cause backend invoke to send stderr to /dev/null instead of merge it with stdout.

dsnopek:
> For earlier versions of Drush, we could add another option that would cause backend invoke to send stderr to /dev/null instead of merge it with stdout.

Ooo, that would be awesome. There's other ways to solve the bug here (ie. always having drush-behat-endpoint return a JSON object, as opposed to other JSON types, or my hacky PR #92) but that would solve several other unrelated annoyances too! :-)

PR #92, opened 2016-06-17
Fix setting specific fields when creating a node via the Drush driver

dsnopek:
Fixes problem caused by #91


jhedstrom:
Thanks for digging into this! I'm not able to test today, but will do so on Monday.

jhedstrom:
Thanks again! This indeed fixes the issue!

dsnopek:
Woohoo! :-)

PR #96, opened 2016-07-22
Support entity links with LinkHandler

### LINKED ISSUE #95: [D8] Support internal links with LinkHandler
The link field in drupal 8 core supports linking to internal pages and is done via an autocomplete link, saved as an entity:node/{nid} url.

The LinkHandler class doesn't support this. It should be similar to how EntityReferenceHandler uses the label key and searches for an entity and replaces the value if so.

Currently attempting this will result in an error on the page:

`InvalidArgumentException: The URI 'Test 123' is invalid. You must use a valid URI scheme. in Drupal\Core\Url::fromUri() (line 281 of core/lib/Drupal/Core/Url.php).`

acbramley:
Fixes #95 

Really crude solution at the moment but just wanted a starting point.


acbramley:
Closing to regenerate PR patch.

PR #98, opened 2016-07-22
Fixed the fieldInfo property name and added datestamp handler

pheudo:


brockfanning:
I hadn't realized it before, but this PR addresses 2 exact issues that I've submitted separate PR's for. My separate PR's are: #108 and #105. Great minds think alike. ;) In the interests of keeping the PR list manageable, we should either close this one, or both of mine. I believe my date fix is a bit more thorough because it adds support for both "Date" and "Datestamp" date fields. @pheudo do you think you could try out that approach, and if it works for you, close this PR?

PR #101, opened 2016-08-07
Create & Delete functions for EntityContext

jonathanjfshaw:
See https://github.com/jhedstrom/drupalextension/issues/194

### DIFF
--- a/src/Drupal/Driver/Cores/CoreInterface.php
+++ b/src/Drupal/Driver/Cores/CoreInterface.php
@@ -231,11 +231,6 @@ interface CoreInterface {
 
   /**
    * Delete an entity.
-   *
-   * @param string $entity_type
-   *   Entity type machine name.
-   * @param $entity
-   *   The entity to delete, as an object with a property 'id' or a true entity.
    */
   public function entityDelete($entity_type, $entity);
 
--- a/src/Drupal/Driver/Cores/CoreInterface.php
+++ b/src/Drupal/Driver/Cores/CoreInterface.php
@@ -218,11 +218,24 @@ interface CoreInterface {
 
   /**
    * Create an entity.
+   *
+   * @param string $entity_type
+   *   Entity type machine name.
+   * @param array $entity
+   *   The field values and properties desired for the new entity.
+   *
+   * @return \Drupal\Core\Entity\EntityInterface
+   *   A new entity object.
    */
   public function entityCreate($entity_type, $entity);
 
   /**
    * Delete an entity.
+   *
+   * @param string $entity_type
+   *   Entity type machine name.
+   * @param $entity
+   *   The entity to delete, as an object with a property 'id' or a true entity.
    */
   public function entityDelete($entity_type, $entity);
 
--- a/src/Drupal/Driver/Cores/CoreInterface.php
+++ b/src/Drupal/Driver/Cores/CoreInterface.php
@@ -216,4 +216,14 @@ interface CoreInterface {
    */
   public function configSet($name, $key, $value);
 
+  /**
+   * Create an entity.
+   */
+  public function entityCreate($entity_type, $entity);
+
+  /**
+   * Delete an entity.
+   */
+  public function entityDelete($entity_type, $entity);
+
 }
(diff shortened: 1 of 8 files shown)


src/Drupal/Driver/Cores/Drupal6.php
```diff
@@ -506,4 +506,18 @@ public function clearStaticCaches() {
     // Drupal 6 doesn't have a way of clearing all static caches.
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function entityCreate($entity_type, $entity) {
+    // Drupal 6 doesn't have an entity API.
```
jhedstrom: This should instead probably be added to `BaseDriver` where it throws an exception instead of silently failing.

src/Drupal/Driver/Cores/Drupal6.php
```diff
@@ -506,4 +506,18 @@ public function clearStaticCaches() {
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function entityDelete($entity_type, $entity) {
+    // Drupal 6 doesn't have an entity API.
```
jhedstrom: Ditto here.

src/Drupal/Driver/Cores/Drupal7.php
```diff
@@ -482,4 +482,18 @@ public function clearStaticCaches() {
     drupal_static_reset();
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function entityCreate($entity_type, $entity) {
+    // @todo: create a D7 version of this function
```
jhedstrom: And here too. Alternatively, since D7 *does* support entities, throw the exception here and note it is a todo item.

src/Drupal/Driver/Cores/Drupal7.php
```diff
@@ -482,4 +482,18 @@ public function clearStaticCaches() {
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function entityDelete($entity_type, $entity) {
+    // @todo: create a D7 version of this function
```
jhedstrom: And here :)

src/Drupal/Driver/Fields/Drupal8/FileHandler.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Drupal\Driver\Fields\Drupal8;
+
+/**
+ * File field handler for Drupal 8.
+ */
+class FileHandler extends EntityReferenceHandler {
```
jhedstrom: Is this part of the necessary changes, or did it sneak in from another PR?

src/Drupal/Driver/Cores/CoreInterface.php
```diff
@@ -216,4 +216,14 @@ public function configGet($name, $key = '');
...
 
+  /**
+   * Create an entity.
+   */
+  public function entityCreate($entity_type, $entity);
+
+  /**
+   * Delete an entity.
+   */
+  public function entityDelete($entity_type, $entity);
```
jhedstrom: These need `@param` docs.

src/Drupal/Driver/Fields/Drupal8/FileHandler.php
jonathanjfshaw: It is not strictly necessary, but that is all that is needed to support file fields now that we have entity support. I will split it into its own PR if you prefer.

jhedstrom: No, it can stay in this PR, I was just curious.

jonathanjfshaw:
I combined both approaches to failures for D6 and D7: I added an exception to the BaseDriver (so that Drush and Blackbox give the right response), and exceptions to D6 /D7 Core core saying Entity API not / not yet supported

src/Drupal/Driver/Cores/CoreInterface.php
jonathanjfshaw: I don't see how to @param the delete function. The $entity argument can be of a number of different types, but code sniffer insists I specify a particular type and use it, so the tests fails if I don't. I tried:
   * @param string $entity_type
   *   Entity type machine name.
   * @param $entity
   *   The entity to delete, as an object with a property 'id' or a true entity.

jhedstrom: I hadn't realized most other methods in here were missing the `@param` as well, so I think what you have is good to go.

jhedstrom:
Thanks!

PR #105, opened 2016-09-07
Two additional date field handlers for D7

brockfanning:
I've run into D7 sites that use other "types" of date fields, and so these are 2 additional classes to simply inherit, without changing, the existing DatetimeHandler class.

### DIFF
--- a/src/Drupal/Driver/Fields/Drupal7/AbstractDateHandler.php
+++ b/src/Drupal/Driver/Fields/Drupal7/AbstractDateHandler.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace Drupal\Driver\Fields\Drupal7;
+
+/**
+ * A base handler for the various types of Date fields in Drupal 7.
+ */
+abstract class AbstractDateHandler extends AbstractHandler {
+
+  /**
+   * Get the format in which the dates are saved in Drupal's database.
+   *
+   * @return string
+   *   The format in which the dates are saved in Drupal's database.
+   */
+  abstract protected function getDateFormat();
+
+  /**
+   * Converts a date string into the format expected by Drupal.
+   *
+   * @return string
+   *   The re-formatted date string.
+   */
+  protected function formatDateValue($value) {
+
+    $date = new DateTime($value);
+    return $date->format($this->getDateFormat());
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function expand($values) {
+
+    $return = array();
+    if (isset($this->fieldInfo['columns']['value2'])) {
+      foreach ($values as $value) {
+        $return[$this->language][] = array(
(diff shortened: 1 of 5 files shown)


jhedstrom:
@brockfanning could you clarify how these would be used?

brockfanning:
@jhedstrom sure thing. Since I wasn't aware of a standard format for dates in Behat tests, my thinking was that the person writing the tests would need to know what format would work for the particular date field being tested, and use that format in the test values. For example, if you know that the field you're testing is a datestamp date field, then the date field value you use in the test should be a Unix timestamp. So these news classes simply allow Drupal 7 date fields of these other types ("date" and "datestamp") to function, in the same way that "datetime" date fields currently work.

You may prefer that the person writing the tests does not need to be aware of the expect date format, and can just use a standard format. In that case, then maybe the format that the DatetimeHandler expects ("Y-m-d h:m:s") should be accepted as that standard format, since it's what everyone has been using thus far. If you think so, I'd be happy to update this PR so that these new classes expect that format, and do their own conversions. Specifically, the DateHandler class would expect "Y-m-d h:m:s" and convert it to "Y-m-d\Th:m:s", and the DatestampHandler class would expect "Y-m-d h:m:s" and convert it to "U".

jhedstrom:
> You may prefer that the person writing the tests does not need to be aware of the expect date format

That would indeed be preferable =)

I guess I'm missing how these 2 empty classes achieve this though? Could you paste sample code that utilizes them?

brockfanning:
Ok great, I'll make those changes and update the PR.

As for how the classes are used, that actually happens automatically, just by virtue of being in that folder and named according to the field type. This, I believe, is the relevant code that handles that: https://github.com/jhedstrom/DrupalDriver/blob/master/src/Drupal/Driver/Cores/AbstractCore.php#L62

brockfanning:
@jhedstrom Ok, this is ready for review again. I added new tests as well.

Also instead of expecting a particular date format in the Behat tests, it works with any date string covered here: http://php.net/manual/en/datetime.formats.php

So users can do something like:

```
| field_my_date |
| +1 month      |
```
or
```
| field_my_date |
| June 2008     |
```
etc...

brockfanning:
Added one more improvement - an "is_array" check to allow for cases where a date field has an end date that is optional, and the Behat test only uses the start date. This change prevents the driver from treating a string value like an array in that case.

AlexSkrypnyk:
Hey @brockfanning - thank you for this contribution! Drupal 7 has reached end of life and we will be dropping D7 support in DrupalDriver 3.x (see #312). Since this PR targets D7 date field handlers only, I'm closing it. If this functionality is still needed for Drupal 10/11, feel free to open a new issue. Appreciate the work!

PR #107, opened 2016-09-16
[D8] Support LinkHandler options

aoturoa:
Currently the LinkHandler disregards and input for the options since it sets a empty array().
This pull request will use the $value[2], split similarly to the Title and Uri, as the options array.

To supply the link options in the behat feature, your input value should look like:
"Link title - http://urihere.com - attributes[target]=_blank"


jhedstrom:
Thanks!

PR #108, opened 2016-09-19
Change nonexistent property "field_info" to "fieldInfo" in taxonomy term reference field handler

brockfanning:
The property name "field_info" isn't right, which is preventing getVocab() from succeeding, in the Drupal 7 taxonomy term reference field handler.


brockfanning:
Just a note, I believe this issue was erroneously referenced, so there is a but of unnecessary noise here. It's actually a pretty simple no-brainer fix.

pfrenssen:
Thanks, good find!

PR #55, opened 2015-08-29
Support allowed_values_function field setting in ListTextHandler:expand()

nedjo:
Currently the Drupal 7 `expand()` method in `ListTextHandler` examines the values in allowed_values. If a field has an `allowed_values_function` set, an incorrect list of allowed values is examined.

There doesn't appear to be equivalent Drupal 8 code.

Commit f879b5e addresses this issue by drawing allowed values from an `allowed_values_function`, if set.

Commit 6b003c1 roughs in possible testing (untested!).

PR #109, opened 2016-09-21
Corrects handling of callback-based lists.

aronbeal:
We looked over this for awhile before we decided it was a bug.  In the case of allowed values in field info, the function had inverted the key/value pairings, and set the *key* as the return value, but in the case of callback functions, it sets the *value* of allowed_values as the return value.  This is despite the fact that the callback function is supposed to return a mapping of raw=>display values, which would mean that the display value was what was stored in the db (which was what was happening for us during api calls).

This corrects the handling so that, in each case, the options array is generated from a flip of the allowed values, and then handling of that options array is identical from that point forward.

### DIFF
--- a/src/Drupal/Driver/DrupalDriver.php
+++ b/src/Drupal/Driver/DrupalDriver.php
@@ -42,7 +42,7 @@ class DrupalDriver implements DriverInterface, SubDriverFinderInterface {
   /**
    * Drupal core version.
    *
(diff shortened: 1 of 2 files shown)


aronbeal:
Related issue: https://github.com/jhedstrom/DrupalDriver/pull/55

jhedstrom:
Thanks!

PR #113, opened 2016-10-13
Add D7 entity create/delete support.

djdevin:
I updated the D7 core stubs to create and delete entities.


jhedstrom:
Sorry for the long delay in review. Thanks for this!

PR #114, opened 2016-11-07
Adding expandEntityBaseFields() method.

grasmash:
This pull request allows expansion of base fields on entities. In Drupal 8, base fields use the same field types as user-defined fields. E.g., a custom entity may define an entity reference field as a base field. Without this PR, there is no mechanism for expanding such fields.

Example usage:
```
      /** @var Drupal8 $d8_core */
      $d8_core = $this->getDriver()->getCore();
      $entity = (object) $row;
      $base_fields = [
        'licensed_entity',
        'user_id',
      ];
      $d8_core->expandEntityBaseFields('license', $entity, $base_fields);
      $d8_core->entityCreate('license', $entity);
```

This would expand the licensed_entity and user_id base fields using the EntityReferenceHandler class.

This pull request also documents a couple of undocumented parameters and fixes a bug that prevented user entity values from being expanded on entity reference fields.

### DIFF
--- a/src/Drupal/Driver/Cores/Drupal8.php
+++ b/src/Drupal/Driver/Cores/Drupal8.php
@@ -375,7 +375,7 @@ class Drupal8 extends AbstractCore {
   /**
    * {@inheritdoc}
    */
-  public function getEntityFieldTypes($entity_type, $base_fields = array()) {
+  public function getEntityFieldTypes($entity_type, array $base_fields = array()) {
     $return = array();
     $fields = \Drupal::entityManager()->getFieldStorageDefinitions($entity_type);
     foreach ($fields as $field_name => $field) {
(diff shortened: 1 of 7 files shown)


src/Drupal/Driver/Cores/Drupal8.php
```diff
@@ -349,13 +350,28 @@ public function getExtensionPathList() {
...
+   * Expands specified base fields on the given entity object to the expected structure.
+   *
+   * @param string $entity_type
+   *   The entity type for which to return the field types.
+   * @param \stdClass $entity
+   *   Entity object.
+   * @param array $base_fields
+   *   Base fields to be expanded in addition to user defined fields.
+   */
+  public function expandEntityBaseFields($entity_type, \stdClass $entity, $base_fields) {
```
grasmash: This new method is public. `expandEntityFields()` cannot be used for custom entities because it is protected.

jhedstrom:
Sorry for the delay. This is merged now. Thanks!

PR #115, opened 2016-12-06
Don't force status for new created nodes

sandervd:
When Node::create($values) is called without status being set in the values,
the node will be created with defaults set in the publishing options on the bundle.

The current implementation forces the node to published, although this might not be intended.
It looks like the code responsible for this is D7 legacy.


pfrenssen:
I had a look at this, and this indeed does not seem to be intentional and probably legacy code. When creating a node type there are a number of publishing options enabled by default, and "Published" is one of them.

So in case a node type is configured to be unpublished by default this won't work as expected, the node will be created as published instead. I think this makes sense that this should respect the default configuration.

It might affect a few people though who are creating nodes are currently relying on the fact that their nodes are magically published when being created. Probably this is a small minority though, and it can be fixed by simply setting the status to `TRUE` before creating the node.

Thanks!

sandervd:
Great, thanks!

PR #116, opened 2016-12-13
Support node links with LinkHandler

### LINKED ISSUE #95: [D8] Support internal links with LinkHandler

acbramley:
Fixes #95

### DIFF
--- a/src/Drupal/Driver/Fields/Drupal8/LinkHandler.php
+++ b/src/Drupal/Driver/Fields/Drupal8/LinkHandler.php
@@ -13,14 +13,29 @@ class LinkHandler extends AbstractHandler {
   public function expand($values) {
     $return = array();
     foreach ($values as $value) {
+      $uri = $value[1];
+
+      // If the uri has no scheme (and is not protocol relative) attempt to find
+      // a node with that label.
+      if (empty($uri_parts['scheme']) && strpos($uri, '//') !== 0) {
+        $entity_type_id = 'node';
+        $query = \Drupal::entityQuery($entity_type_id)->condition('title', $uri);
+        $entities = $query->execute();
+        if (!empty($entities)) {
+          $nid = array_shift($entities);
+          $uri = 'entity:' . $entity_type_id . '/' . $nid;
+        }
+      }
+
       $return[] = array(
         // 'options' is required to be an array, otherwise the utility class
         // Drupal\Core\Utility\UnroutedUrlAssembler::assemble() will complain.
         'options' => array(),
         'title' => $value[0],
-        'uri' => $value[1],
+        'uri' => $uri,
       );
     }
+
     return $return;
   }
 


src/Drupal/Driver/Fields/Drupal8/LinkHandler.php
```diff
@@ -13,18 +13,34 @@ class LinkHandler extends AbstractHandler {
...
+      // a node with that label.
+      if (empty($uri_parts['scheme']) && strpos($uri, '//') !== 0) {
+        $entity_type_id = 'node';
+        $query = \Drupal::entityQuery($entity_type_id)->condition('title', $uri);
+        $entities = $query->execute();
+        if (!empty($entities)) {
+          $nid = array_shift($entities);
+          $uri = 'entity:' . $entity_type_id . '/' . $nid;
+        }
+      }
```
pfrenssen: I don't think this is the right way to approach this. The goal of DrupalDriver is to abstract the Drupal API, not the user interface.

When creating link fields through the API it is also not possible to supply a node title for the link field value and saving the entity like this. The resolving of the titles is a UI feature that is handled in the autocomplete. Behind the scenes the entity will be saved using the resolved link, not using the title that was originally entered in the UI.

 What would be interesting though is to support internal URIs, since these are also supported by the API. The `link_type` property on the link field controls whether or not internal links are accepted.

acbramley: I agree! This is bringing the LinkHandler in line with what the EntityReferenceHandler does which let's you reference entities by their label.

Using internal URIs would work with the current implementation however how would you know what the entity_id is of an entity you just created in your behat test?

pfrenssen:
If this is about Behat testing then a better place for discussing this is in the DrupalExtension :) DrupalDriver is just a dependency of DrupalExtension, it is not part of it. The goal of DrupalDriver is to abstract the Drupal API, so you can execute the same code against different versions of Drupal with the same end result. This is very handy for DrupalExtension since it provides Behat integration for Drupal 6, 7 and 8.

To explain more clearly what the use case of DrupalDriver is and how it differs from DrupalExtension:

When you create a node in a Behat test. your Behat test code will get the data from your test scenario and at some point it wants to instruct Drupal to save the node in the database. It basically wants to call `node_save()` but it wants to do this in a version-agnostic way. So instead it calls `DrupalDriver::getCore()->nodeCreate()`. DrupalDriver will then take the data, and transform it in the format that is expected by the actual Drupal version you are running, and then call the respective function. On D6 this will be `node_save()`, on D8 `Node::create()`.

What `expand()` is doing is to prepare the version agnostic data to version specific data. There are differences between the different Drupal versions. For example the values for a link field differ between D7 and D8:

```
// D7 expanded link field value:
$values['field_link'][LANGUAGE_NONE][0] = [
  'title' => $title,
  'url' => $url,
];

// D8 expanded link field value:
$values['field_link'][0] => [
  'title' => $title,
  'uri' => $url,
];
```

As you can see in D7 we have `LANGUAGE_NONE` in the array, and the link is called 'url' instead of 'uri'. So it is expanding the data into the exact format that is expected by the Drupal API.

Transforming labels into IDs is not handled by the Drupal API either, you cannot call this:

```
$node = Node::create([
  'field_link' => ['uri' => 'Some title of an entity I want to reference'],
]);
```

This stuff is handled in the form submission code. In the end the resolved URI will be passed to `Node::create()`, this is what I meant with the UI code.

Transforming human readable values to machine IDs is firmly in the realm of Behat testing (i.e. DrupalExtension). You should put it somewhere in your test code, like in `FeatureContext`.

Now to get you up and running quickly, I have some example code for you, but I'll put it in a DrupalExtension issue, so that it will be useful too for other people that are looking for this information.

acbramley:
Thanks for the thorough description but I already understand what you've described. Please look at the EntityReferenceHandler to see what I'm referring to (this is in DrupalDriver).

That is the handler for entity reference fields and transforms entity labels into the proper format for Drupal (i.e a target_id). By your own argument this isn't supported by the Drupal API either since this doens't work either:

`
$node = Node::create([
  'field_entity_ref' => ['target_id' => 'Some title of an entity I want to reference'],
]);
`

So why is it in the driver? In fact the handler explicitly does not support straight entity ids!

I understand where you're coming from but I'm simply following patterns that are already set in this library.

pfrenssen:
Hmm yeah you're right, It is abusing the expand method for a Behat use case. This doesn't belong there. We cannot remove that any more though because it would break backwards compatibility :(

AlexSkrypnyk:
Hey @acbramley - thank you for this contribution! LinkHandler has been updated in #311 to support named field properties. I'm closing this as superseded. Appreciate the work!

PR #117, opened 2017-01-02
Fix d8 User Entity Reference fields

Sut3kh:
Creating a node with an user entity reference field value throws:
```
'Drupal\Core\Entity\Query\QueryException' with message ''' not found' in web/core/lib/Drupal/Core/Entity/Query/Sql/Tables.php:316
```

It would seem that the User entity does not define the label column, an issue that the d7 EntityReferenceHandler appears to work around:
```
// For users set label to username.
if ($entity_type == 'user') {
  $entity_info['entity keys']['label'] = 'name';
}
```

So i have used the same approach to fix it, manually setting the label_key for user entities.

### DIFF
--- a/src/Drupal/Driver/Fields/Drupal8/EntityReferenceHandler.php
+++ b/src/Drupal/Driver/Fields/Drupal8/EntityReferenceHandler.php
@@ -14,7 +14,15 @@ class EntityReferenceHandler extends AbstractHandler {
     $return = array();
     $entity_type_id = $this->fieldInfo->getSetting('target_type');
     $entity_definition = \Drupal::entityManager()->getDefinition($entity_type_id);
-    $label_key = $entity_definition->getKey('label');
+
+    // Determine label field key.
+    if ($entity_type_id !== 'user') {
+      $label_key = $entity_definition->getKey('label');
+    }
+    else {
+      // Entity Definition->getKey('label') returns false for users.
+      $label_key = 'name';
+    }
 
     // Determine target bundle restrictions.
     $target_bundle_key = NULL;


jonathanjfshaw:
https://www.drupal.org/node/2448729 is the relevant Drupal issue, which holds no help at present.

Hardcoding a reference to User is horrible but easy. The only alternative I can see without rearchitecting the whole thing is to iterate over the entity base fields and choose the first one that is required but not the id or uuid key. 

However,  that seems like not much of an improvement, so I suggest we do it your way for now.

pfrenssen:
I don't like the solution either, but we are already doing the same thing for D7 so this will do for now :)

jonathanjfshaw:
Agreed it's good enough for now. I think it is something that Robust Entity handling (https://github.com/jhedstrom/drupalextension/issues/337) might be able to help with. Basically if we had something like our own Behat Drupal User class (extending a Behat Drupal Entity class) that offered the properties/methods we needed. Then we could query the labelField() method of the class and get the field name we needed.

jonathanjfshaw:
The CS issues that cause the test fails are unrelated, and I think have been fixed since the last tests ran here.

pwolanin:
I just ran into this exact bug an applied the same fix - any ETA?

Looking at https://www.drupal.org/node/2448729 there are several possible problem entity types - maybe need a switch statement to handle the ones in core at least?

pfrenssen:
Oh yeah we should fix this. I wish github had some kind of way to follow or flag issues, I have no time to look at it now, and when I will have time I won't be able to find it any more :(

jonathanjfshaw:
As a maintainer you've got access to issue labels;
using a few like "Architecture", "Priorities" and "RTBC" might help.
You can filter issues by label.

I'd be up for helping with issue triage if you like, but alas it requires
giving me write access(though I'd very happily agree never to commit to
master or merge PR's).

Useful things I could do with that:
- label issues
- close ancient dead issues
- flag issues that need maintainer attention
- trigger travis retests

pwolanin:
The other entity types are less important than User - would be great to get this in.

pwolanin:
Note that this is working for me (applied as a patch in composer.json and testing a node type with a user entity reference field)

jonathanjfshaw:
Working for me too.

jonathanjfshaw:
See also #94 and #133 for discussions addressing this problem

Berdir:
Bump :)

Wouldn't hurt to at least add a check to make sure there really is a label but agreed, user is by far the most common use case.

jhedstrom:
I'm fine with this once #139 is resolved so we can verify tests are passing here.

jhedstrom:
Tests are green again. I've kicked off a rebuild on travis.

jhedstrom:
I think this needs to be rebased off of the latest `master`. I can't tell why the hhvm tests are failing here but not there.

Sut3kh:
done, sorry for the delay

jhedstrom:
Thanks everybody!

PR #120, opened 2017-01-03
Speed up test runs by being satisfied with release packages

pfrenssen:
Currently we're using the `--prefer-source` option to download our dependencies. This is slow since the complete git repositories will be downloaded for every dependency. I don't think we have an actual need for the full source packages of our dependencies, so we can download the faster distribution builds instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ php:
   - hhvm
 
 install:
-  - composer install --dev --prefer-source
+  - composer install
 
 script:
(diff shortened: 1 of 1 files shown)


pfrenssen:
Mission accomplished!

* Before: 11 min 17 sec (ref. [build 453|https://travis-ci.org/jhedstrom/DrupalDriver/builds/188532871])
* After: 7 min 37 sec (ref. [build 454|https://travis-ci.org/jhedstrom/DrupalDriver/builds/188533925])

Total speedup: 32.5% :+1:

jhedstrom:
👍

Issue #122, opened 2017-01-20
Missing code in FileHandler for Drupal 8 field expand


azurams:
I'm trying to create an entity with a file field, unfortunately it fails.
Whereas quite the same works for an entity with an "image field"

AlexSkrypnyk:
Resolved by #314.

PR #123, opened 2017-01-20
handles File field expansion for D8 codebase

azurams:
this is related to the issue #122


AlexSkrypnyk:
Hey @azurams - thanks for this contribution! It's been a while since the last update here, so I've picked up the work and iterated on it. I've opened #314 which re-implements the FileHandler using the modern file.repository service and adds support for multi-value files and named keys. Your contribution is attributed there. I'll close this PR once the new one is merged. Appreciate the initial work!

PR #125, opened 2017-02-07
Load drushrc.php for D8.

TwoD:
Both D6 and D7 cores load the drushrc file if they find it, but the D8 core didn't. Without this we could not easily run tests on multisites created by Aegir.


ashleyearl:
I'm currently patching this fix in before our test runs as it resolves issues we otherwise have running tests against sites configured using aegir. The fix looks good to me. Is there a reason it hasn't been merged in yet? It would be nice to not have to manually apply a patch before running our tests.

pfrenssen:
Looks good, this improves the consistency between the 3 supported Drupal versions. Thanks!

PR #126, opened 2017-03-24
Infer timezone in DateTime field handler

jonathanjfshaw:
I thought I had already made PR for this but I can't find it, forgive me if I'm being foolish.

Currently the following scenario will fail most of the time:
```
Given I am viewing an article:
  | title                  | test                            |
  | field_datetime | 2005-06-30 14:00:00 |
Then I should see "2005-06-30 14:00:00"
```

It will typically fail because values are by default created and stored using UTC, but rendered using the current user's timezone. The workaround is for step authors to do mental time math, writing their "Given" times in UTC and their "Then" times in the Drupal timzeone (which is the default for any user).

I think we should assume that step authors intend time values to be in the Drupal default timezone when they write them in Given statements, unless they explicitly specify a timezone.

This will break BC for anyone writing tortured scenarios expecting the current behavior.

I can write some tests for it in the extension, if you like the idea of this.

### DIFF
--- a/src/Drupal/Driver/Fields/Drupal8/DatetimeHandler.php
+++ b/src/Drupal/Driver/Fields/Drupal8/DatetimeHandler.php
@@ -2,6 +2,9 @@
 
 namespace Drupal\Driver\Fields\Drupal8;
 
+use DateTime;
+use DateTimeZone;
+
 /**
  * Datetime field handler for Drupal 8.
  */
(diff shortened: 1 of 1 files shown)


jonathanjfshaw:
We may want to ensure forwards compatibility with https://www.drupal.org/node/2632040

jonathanjfshaw:
I now know that forward-compatibility with https://www.drupal.org/node/2632040 is not an issue. We could add extra logic for that, but it won't invalidate what I suggest here.

jhedstrom:
Should this go in before or after #167?

jonathanjfshaw:
It doesn't need to wait on #167.

As i said in #167 there is a case for a big picture discussion on date
handling (which would define all the desired features including this one),
and then once that's done making date_range extend date.

But I see no great harm in adding features like this piecemeal, these
handlers are still pretty simple.

jhedstrom:
Makes sense to me. Thanks for this!

bkosborne:
This created an issue in #200

PR #127, opened 2017-04-19
Drupal 8: 'Given :vocabulary terms:' parent doesn't work as documented

mariacha:
From the DrupalExtension:
https://github.com/jhedstrom/drupalextension/blob/master/src/Drupal/DrupalExtension/Context/DrupalContext.php#L378

The example given is:

```
   | name  | parent | description | weight | taxonomy_field_image |
   | Snook | Fish   | Marine fish | 10     | snook-123.jpg        |
```

That theoretically should create a term "Snook" and link it to a term "Fish", which is the parent of Snook.

In Drupal 7, this works:
https://github.com/jhedstrom/DrupalDriver/blob/master/src/Drupal/Driver/Cores/Drupal7.php#L306

But in Drupal 8, there's no such handling for the parent, so you have to add the parent's ID. 
https://github.com/jhedstrom/DrupalDriver/blob/master/src/Drupal/Driver/Cores/Drupal8.php#L310

This adds the parent lookup code.

### DIFF
--- a/src/Drupal/Driver/Cores/Drupal8.php
+++ b/src/Drupal/Driver/Cores/Drupal8.php
@@ -314,7 +314,7 @@ class Drupal8 extends AbstractCore {
       $parent = \taxonomy_term_load_multiple_by_name($term->parent, $term->vocabulary_machine_name);
       if (!empty($parent)) {
         $parent = reset($parent);
-        $term->parent = $parent->tid;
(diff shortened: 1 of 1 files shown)


mariacha:
Hey hey! Build passes now :)

jhedstrom:
Thanks!

PR #130, opened 2017-04-25
Entity reference field handler: allow force target id

omarlopesino:
Hi.

Now the step 'Given :type content' doesn't allow force target id for entity reference fields, in this way:
```
 | field_entityrereference:target_id |
 | 123                                              |
```

I need this in a project where there is a entity reference field which points to a remote entity. That entity doesn't have a single label column (it needs 2 to build the label) and it's needed force target id.

Can you review this pull request, please? Thanks!


AlexSkrypnyk:
Hey @omarlopesino - thank you for this contribution! Drupal 7 has reached end of life and we will be dropping D7 support in DrupalDriver 3.x (see #312). Since this PR targets D7 field handlers only, I'm closing it. If this functionality is still needed for Drupal 10/11, feel free to open a new issue. Appreciate the work!

PR #134, opened 2017-07-13
Support for testing emails sent from Drupal

jonathanjfshaw:


jonathanjfshaw:
For use in parallel drupal-extension PR https://github.com/jhedstrom/drupalextension/pull/392

jonathanjfshaw:
Updated PR to use config override to enable the test mail collector, rather than unnecessary and risky modification of stored config.

jhedstrom:
I've rebased and merged this in #171. Thanks for all the work!

PR #137, opened 2017-09-22
Allow link fields to have only URL

ohthehugemanatee:
The current D8 LinkHandler assumes an array input, which makes testing with an empty link "title" field impossible. This PR tests the value first to see if it's an array. If not, it assumes the value is just the URL and to leave the other values out.


AlexSkrypnyk:
Hey @ohthehugemanatee - thank you for this contribution! URI-only link support is being landed via an iteration on #267 which builds on the same idea. Your contribution will be credited there. Appreciate the initial work!

PR #151, opened 2017-11-22
Fix illegal string offset warning.

### LINKED ISSUE #150: isLegacyDrush() iillegal offset 'drush-version'
I have a bamboo based build system that is reporting failed Behat tests due to:

Warning: Illegal string offset 'drush-version' in vendor/drupal/drupal-driver/src/Drupal/Driver/DrushDriver.php line 139

Ive since updated my local version to the latest as well but dont see the issue when running the tests locally.

Running on Acquia btw if that has any impact. Running Drupal 7 on Drush 8.

Anyone got any ideas why this would be the case?

EDIT - it seems Acquia have removed part of this in some of their Drupal 8 build tool stuff:

https://github.com/acquia/blt/commit/1f5419b84bcaa90fe83f7004b4cf5664405a5df5

Is the code used in isLegacyDrush correct based on the above?

jhedstrom:
- Fixes #150

### DIFF
--- a/src/Drupal/Driver/DrushDriver.php
+++ b/src/Drupal/Driver/DrushDriver.php
@@ -135,10 +135,12 @@ class DrushDriver extends BaseDriver {
   protected function isLegacyDrush() {
     try {
       // Try for a drush 9 version.
-      $version = unserialize($this->drush('version', [], ['format' => 'php']));
-      return version_compare($version['drush-version'], '9', '<=');
+      $version = trim($this->drush('version', [], ['format' => 'string']));
+      return version_compare($version, '9', '<=');
     }
     catch (\RuntimeException $e) {
+      // The version of drush is old enough that only `--version` was available,
+      // so this is a legacy version.
       return TRUE;
     }
   }


greg-1-anderson:
Current HEAD of the Drush 8.x branch is behaving like this:
```
$ drush version --format=php
s:10:"8.1.16-dev";
```
Drush 9 is still doing this:
```
$ drush9 version --format=php
a:1:{s:13:"drush-version";s:17:"9.0-dev-gbfd873ad";}
```

jhedstrom:
Thanks @greg-1-anderson, I'll take that into account here.

greg-1-anderson:
Actually, every version of Drush 8 and Drush 7 that I checked returns a simple string for `drush version`. Maybe we should try to get Drush 9 to conform.

In any event, accepting either a string or an array here would be the defensive way to go.

greg-1-anderson:
Actually, I think that we'll probably keep Drush 9 as it is. I'd recommend that you use `drush version --format=string`. That behaves consistently between Drush 8 and Drush 9, and you don't need to worry about parsing or array key values.

PR #152, opened 2017-11-25
Node reference handler for Drupal7

marcelovani:
I have a situation where more than 20 sites use _references_ module.
I can't figure out how to make the Drupal driver use this code unless it is in the specific folder.
Would it be a bad idea to have this in the drupal driver codebase?


AlexSkrypnyk:
Hey @marcelovani - thank you for this contribution! Drupal 7 has reached end of life and we will be dropping D7 support in DrupalDriver 3.x (see #312). Since this PR adds a D7-only node reference handler, I'm closing it. If this functionality is still needed for Drupal 10/11, feel free to open a new issue. Appreciate the work!

Issue #155, opened 2017-12-28
Entity & Field plugins


jonathanjfshaw:
I'm most of the way there with my Christmas present: an initial rough implementation of

1.  Field plugins (real D8 plugins that developer can supplement for their own project); these replace the home-baked Field handlers.
2.  Entity  plugins (rationalising up all the node, term, user, etc. creation & deletion logic and allowing developers to override/supplement it.
3.  Extensive driver-side kernel tests to make sure everything works the same after the refactoring. 

I'm thinking this should be for version 2.0; do you have a timeline for that or other things on the roadmap for it?

jonathanjfshaw:
Initial discussion about this happened a long time ago on the extension-side at https://github.com/jhedstrom/drupalextension/issues/337

jhedstrom:
This sounds great. I'm not sure if it would be the same plugin system or not, but I'd like to split the individual drivers up into collections of plugins so they aren't these singular monolithic classes (I'm not sure of how that would work in practice).

jonathanjfshaw:
Entity and field plugins should serve to remove about 12+ methods from the
drivers, so they're a step on this process of taming the monolith. I have
wondered if the drivers should be collections of services, with the entity
and field plugin managers as two of those services.

jonathanjfshaw:
I've completed initial development of an entity and field plugin/wrapper system, see PR #157. 

## Benefits

*Customisation*: users can add special handling for new field or entity types, or override handling of standard types.

*Simplicity*: we can get rid of all entity-type specific methods in the driver and extension. We can even throw away the whole hook system in the extension.

*Labels*: identifying entity types, bundles, fields and referenced entities by labels rather than machine names is (more) possible and customisable.

*Cleaning up*: we can specify additional tear down instructions for particular entity types.


## Overview

In a nutshell if you want to create a new Drupal entity using the driver what now happens is:
* you ask the driver for a new DriverEntity wrapper object and tell it what entity type.
* you tell the DriverEntity wrapper object additional information about the entity (like bundle and fields)
* When it needs to, the DriverEntity wrapper uses the DriverEntityPluginManager to discover and use the most-targeted DriverEntity plugin
* the DriverEntity plugin abstracts in a customisable way a real Drupal entity.
* whenever a user-supplied field value needs to be interpreted the DriverEntity wrapper or plugin calls up the DriverFieldPluginManager and uses that to discover and process the field value through all matching DriverField plugins.

[...]

PR #157, opened 2018-01-14
Entity & Field Plugins

jonathanjfshaw:
Addresses #155. I've added a general discussion of top-level matters to that issue.

Some notes for reviewers on implementation matters...

## Code flow

The code flow typically begins with a DriverEntityDrupal8 wrapper object.

This then discovers and wraps a DriverEntity plugin, and the DriverEntity plugin creates or loads and wraps a drupal entity.

The DriverEntity Plugin calls DriverField wrapper object when it wants to process field values. The DriverField wrapper discovers and process through matching field plugins.

## Review order

I suggest reviewing in this order:

1. Start by looking at the plugins, to see how end developers would work with this stuff.
2. Then look at the Driver's Drupal8 core to see how the driver calls these new systems.
3. Then look at the Entity kernel tests both to get confidence and to see some more API examples.
4. Then look at DriverEntityDrupal8 and it's parent DriverEntityBase. You'll see that this does 3 things:
- gets itself a DriverEntity plugin
- resolves the question of what the bundle is, if and when it can.
- passes almost everything else straight on to the plugin.
5. Then look at DriverEntityPluginDrupal8Base and its parent DriverEntityPluginBase. You'll see that the functionality plugins provide is for 3 main purposes:
- to access the Drupal API and provide information like bundle keys that are needed for plugin discovery etc. to work
- to discover field plugins and process field values through them
- to wrap the real Drupal entity and it's standard methods
6. Next up look at the DriverEntityPluginManager and base to see how plugins are discovered and the right plugin is identified.
7. If you're still alive, look at the DriverField wrapper, plugins, plugin manager and tests.

## The bundle dance

A source of major pain in the DriverEntity wrapper is that when creating an entity we are sometimes explicitly told what bundle it should be, sometimes it has no bundle, and sometimes it has a bundle but we have to find it amongst the supplied fields. But the bundle is part of what is (optionally) used to identify the right plugin for an entity, so we don't get the definitive plugin and a drupal entity until we've resolved the bundle.

My solution is to distinguish between a provisional (bundle-agnostic) plugin loaded on a mere entity-type match, and a final plugin (which may or may not be different) that also uses the bundle when considering which plugin to use. This allows us to use plugin methods even before we have resolved the bundle, provided they are not methods that set data on the plugin that would be lost if we switched to a different final plugin. Basically it allows us to load a provisional plugin matched by entity type, ask that plugin what the bundle key and bundle key labels are for this entity type, and use that information to find the bundle field from amongst an array of bundles; once the bundle is set we hen ask for the definitive plugin with which we do much more.

## The DriverField wrapper

Th DriverField wrapper does wrap a plugin, but it doesn't wrap a Drupal object in the way that the DriverEntity wrapper does. I imagine the DriverEntity wrapper objects will be stored in the Behat extension once instantiated, instead of the repository of ids we currently have, and could be used for all kind of purposes. The DriverField  wrapper is much more of a single trick pony. I do wonder whether it should actually be slimmed altogether and just become a method on the DriverEntity plugin base. The danger is that would fatten up the DriverEntity plugin base, which is already complex.

## ExpandEntityFields ignored base fields

The ExpandEntityFields method on the driver made use of the driver's getEntityTypes and isField. This seems to have made expanding not happen for baseFields. This behaviour is undocumented, and I could see no purpose to it, so I have not continued it.

## ParseEntityFields

The ParseEntityFields method on the extension's RawDrupalContext is a rats nest of untested branches and loops, that frequently calls in to the driver's getEntityTypes and isField methods. It should probably be using a DriverEntity wrapper to call methods on a DriverEntity plugin, and we could thendeprecte the driver's getEntityTypes and isField. But I'm not sure what the plugin methods should be and figuring it out would require a complete refactoring and unit testing of parseEntityFields.

## Tests
There is a failing user kernel test because of the call to validateDrupal in the driver's userCreate, discussed in #155.


jhedstrom:
Wow, this is an amazing start! For resolving the tests, we'll need to pull down Drupal (not sure we need to add it as a dev dependency or not...) I'll try to give this a proper review sometime soon.

haringsrob:
I will try to test this later as it looks promising, for now we were in the need of adding all sorts of entities as well and we came up with this:

```
  /**
   * The entities created.
   *
   * @var \Drupal\Core\Entity\EntityInterface[]
   */
  private $createdEntities = [];

  /**
   * Creates an entity of a given type.
   *
   * @Given :entity with bundle :type content:
   *
   * @throws \Exception
   */
  public function createEntityWithBundle(string $entity_type, string $type, TableNode $table) {
    foreach ($table->getHash() as $entityHash) {
      $entity_data = (object) $entityHash;
      $entity_data->type = $type;

      /** @var \Drupal\Driver\Cores\Drupal8 $core */
      $core = $this->getDrupal()->getDriver()->core;

      $method = new ReflectionMethod(Drupal8::class, 'expandEntityFields');
      $method->setAccessible(TRUE);

      $this->parseEntityFields($entity_type, $entity_data);
      $method->invoke($core, $entity_type, $entity_data);

      $definitions = \Drupal::entityTypeManager()->getDefinitions();
      if (isset($definitions[$entity_type])) {
        /** @var \Drupal\Core\Entity\EntityInterface $class */
        $class = $definitions[$entity_type]->getOriginalClass();
        $entity = $class::create((array) $entity_data);
        if (!$entity->save()) {
          throw new \Exception(sprintf('Entity could not be saved.'));
        }
        $this->createdEntities[] = $entity;
      }
      else {
        throw new \Exception(sprintf('Entity type %entity_type does not exist.', ['%entity_type' => $entity_type]));
      }
    }
  }

  /**
   * Cleanup entities after scenario.
   *
   * @afterScenario
   */
  public function cleanEntities() {
    foreach ($this->createdEntities as $entity) {
      $entity->delete();
    }
  }
```

Example:
```
    And "commerce_product" with bundle "default" content:
      | title |
      | test  |
      | test2 |
```

jonathanjfshaw:
I've got an old PR for that on the extension:
https://github.com/jhedstrom/drupalextension/pull/300

I'm using it as a composer patch, works OK. Doesn't use this new plugin
goodness from the driver side, uses the entity handling yo committed to the
driver a year ago.

jhedstrom:
I've just kicked tests off for the Drupal Extension with this branch to see what happens.

https://travis-ci.org/jhedstrom/drupalextension/builds/374181918

jhedstrom:
Looks like the failures would be resolved by rebasing this to include the mail stuff that went in with #134.

src/Drupal/Component/Utility/DriverNameMatcher.php
```diff
@@ -0,0 +1,272 @@
...
+        $targets = [$target => $target];
+        $results = $this->identifySet($targets);
+        // Get the first key of the results.
+        reset($results);
+        return key($results);
+    }
+
+    /**
+   * Identifies the targets from the pool of candidates.
+   *
```
jhedstrom: Basic indentation fix.

src/Drupal/Component/Utility/DriverNameMatcher.php
```diff
@@ -0,0 +1,272 @@
...
+   * @return array
+   *   For each matched target, the key will be replaced with the machine name
+   *  of the matching candidate, & the value will be preserved. Order may vary.
+   */
+    public function identifySet($targets)
+    {
+        if (is_array($targets)) {
+            $this->targets = $targets;
+        } else {
+            throw new \Exception("Targets to be identified must be passed as an array with their human-friendly name as the keys and anything as the values.");
```
jhedstrom: Is there a reason we can't just type-hint the function? (`array $targets`)

src/Drupal/Component/Utility/DriverNameMatcher.php
```diff
@@ -0,0 +1,272 @@
...
+        $mayHavePrefix = !is_null($this->prefix);
+        $this->identifyByMethod("MachineNameExactly");
+        $this->identifyByMethod("LabelExactly");
+        if ($mayHavePrefix) {
+            $this->identifyByMethod("MachineNameWithoutPrefix");
+        }
+        $this->identifyByMethod("MachineNameWithoutUnderscores");
+        if ($mayHavePrefix) {
+            $this->identifyByMethod("MachineNameWithoutPrefixAndUnderscores");
+        }
```
jhedstrom: The logic above here could use some code comments I think. It would be hard to grok 6-months, or a year, down the road I think.

src/Drupal/Component/Utility/DriverNameMatcher.php
```diff
@@ -0,0 +1,272 @@
...
+                // Skip over candidates that describe fields already matched.
+                if (in_array($machineName, $matchedCandidates)) {
+                    continue;
+                }
+                // If the identification method determines a match, remove the candidate
+                // and target from future consideration, and save the result.
+                if ($this->$methodFunctionName($identifier, $machineName, $label)) {
+                    //$this->candidates = array_filter($this->candidates, function ($value, $key) use ($machineName) {
+                     // return $value === $machineName;
+                    //}, ARRAY_FILTER_USE_BOTH);
```
jhedstrom: This looks like it's been moved down below, so can be removed here?

src/Drupal/Driver/Cores/Drupal8.php
```diff
@@ -54,38 +58,21 @@ public function clearCache() {
...
-    if (isset($node->author)) {
-      $user = user_load_by_name($node->author);
-      if ($user) {
-        $node->uid = $user->id();
-      }
-    }
-    $this->expandEntityFields('node', $node);
-    $entity = Node::create((array) $node);
+    $entity = $this->getNewEntity('node');
+    $entity->setFields((array) $node);
```
jhedstrom: I love how simple this becomes! 👍

src/Drupal/Driver/Cores/Drupal8.php
```diff
@@ -308,20 +214,9 @@ public function validateDrupalSite() {
...
-      if (!empty($parent)) {
-        $parent = reset($parent);
-        $term->parent = $parent->id();
-      }
-    }
-
-    $this->expandEntityFields('taxonomy_term', $term);
-    $entity = Term::create((array) $term);
+    $entity = $this->getNewEntity('taxonomy_term');
+    $entity->setFields((array) $term);
```
jhedstrom: 👍

tests/Drupal/Tests/Driver/DriverNameMatcherTest.php
```diff
@@ -0,0 +1,155 @@
+<?php
+
+namespace Drupal\Tests\Driver;
+
+use Drupal\Component\Utility\DriverNameMatcher;
+
+/**
+ * Tests the Driver's name matcher utility.
+ */
+class DriverNameMatcherTest extends \PHPUnit_Framework_TestCase
```
jhedstrom: Yay unit tests!

jhedstrom:
I still need to review some of the low-level implementation, but getting tests green here, and on the Drupal Extension will definitely be good :)

jhedstrom:
I re-triggered the test run and there appears to be some interface compatibility issues: https://travis-ci.org/jhedstrom/drupalextension/jobs/374181923

jonathanjfshaw:
OK, this time the rebase is good, all green locally except the expected fail.

jonathanjfshaw:
Now kernel tests and extension features passing locally with php5.5 & D8.5.

jonathanjfshaw:
Coding standards fixed. Now green and clean locally, should be green for CS here.

jonathanjfshaw:
My attempts to execute the kernel tests building on #188 have foundered because we have a namespace collision with core's composer.json:
```
    "autoload": {
        "psr-4": {
            "Drupal\\Core\\": "lib/Drupal/Core",
            "Drupal\\Component\\": "lib/Drupal/Component",
            "Drupal\\Driver\\": "../drivers/lib/Drupal/Driver"
        },

```

hence phpunit gives me

```Fatal error: Class 'Drupal\KernelTests\Core\Entity\EntityKernelTestBase' not found in /app/tests/Drupal/Tests/Driver/Kernel/Drupal8/Entity/DriverEntityKernelTestBase.php on line 13```

jonathanjfshaw:
I misdiagnosed the problem, it wasn't namespace collision, it was just that Drupal doesn't  autoload the KernelTests namespace normally, it relies on tests/bootstrap.php to do that.

## PHP 5.5
I have a problem supporting php5.5 because the new DriverNameMatcher class makes heavy use of the $flag argument (ARRAY_FILTER_USE_BOTH) to array_filter introduced in php5.6. 

Additionally Drush 9 (required for Drupal 8.5) doesn't support php5.5, so it's tricky to install Drupal and run kernel tests. The drupal extension shows how to work round this in principal though.

As I understand it Drupal 6 requires php 5.5, so likely I need to rework DriverNameMatcher to work in php 5.5.

## PHP 7.2
PHP 7.2 requires phpunit > 6.0, but we're pinned at ~4.0. Drupal is [working round this awkwardly](https://www.drupal.org/project/drupal/issues/2937469), but we can't call their workaround directly.

## Expected fails & next steps

There's 3 big questions we need to resolve at this point:

1. We have 2 expected test fails, due to the call to validateDrupal in the driver's userCreate, discussed in #155. Why is that call there?

2. It's possible the driver field wrappers should be removed, as discussed in the issue summary above.

3. There's complex things happening in the driver when it comes to base fields and configurable fields. And the driver has made changes in this area this year. See the issue summary again.  I don't know what the desired behavior now is and if I've preserved it (these plugins work with both base and configurable fields without distinction AFAIK).

PR #159, opened 2018-01-20
Ignore access on D8 EntityReferenceHandler

### LINKED ISSUE #158: Don't check access when referencing entities
We should add ->accessCheck(FALSE) to our EntityQuery in the D8 EntityReferenceHandler.

Otherwise it will fail when asked to create entities that reference other entities the current user cannot access. The Driver shouldn't really care about the current user unless explicitly told to, as it's methods are often used to set up context for a test rather than describe actions of a test user.

jonathanjfshaw:
Fixes #158.


jhedstrom:
Thanks!

PR #161, opened 2018-01-30
Daterange handler

### LINKED ISSUE #160: Daterange field handler for Drupal 8
Now that core supports Datetime ranges, we should too.

jonathanjfshaw:
Not at all a great solution, but maybe good enough for now until we get real plugins and kernel tests. Fixes #160.


jhedstrom:
Looks good. Thanks!

PR #180, opened 2018-03-20
Support for the mailsystem module.

### LINKED ISSUE #179: Add support for the mailsystem module
Since the [Mail System](https://www.drupal.org/project/mailsystem) module is necessary for full HTML emails (coupled usually with swiftmailer), it would be good to support that here since the feature added in #134 only supports the core mail manager's approach of sending mail.

jhedstrom:
- Fixes #179


jhedstrom:
This gets things sort of working, but I discovered #181 while working on this.

PR #182, opened 2018-03-20
Persist mail configuration changes to storage.

### LINKED ISSUE #181: Using module override for mail collection doesn't fully work
Utilizing the module overrides for the mail collection doesn't fully work. It only changes the config in the bootstrap that Behat is running from. Thus, any mails sent by filling out forms, or events that happen in a separate bootstrap go back to using the normal mail collector.

Rather than using module override, I think we'll have to actually save the config (and track the original values), then restore them afterward.

jhedstrom:
- Fixes #181

### DIFF
--- a/src/Drupal/Driver/Cores/Drupal8.php
+++ b/src/Drupal/Driver/Cores/Drupal8.php
@@ -21,6 +21,17 @@ use Symfony\Component\HttpFoundation\Request;
 class Drupal8 extends AbstractCore {
 
   /**
+   * Tracks original configuration values.
+   *
+   * This is necessary since configurations modified here are actually saved so
+   * that they persist values across bootstraps.
+   *
+   * @var array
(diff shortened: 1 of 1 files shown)


jonathanjfshaw:
I'm concerned about something here
```
+    $config = \Drupal::configFactory()->getEditable('system.mail');
+    $data = $config->getRawData();
+
+    // Save the original values for restoration after.
+    $this->originalConfiguration['system.mail'] = $data;
```

If startCollectingMail is called twice before stopCollectingMail is called (and I think this might happen in the parrallel behat extension PR), then test_mail_collector will end up being stored in $originalConfiguration and persisted. 

Maybe we need to wrap the above hunk in 
`if (!isset($this->originalConfiguration['system.mail'])) {}`

jhedstrom:
Yeah, I think that change would make sense.

Issue #183, opened 2018-04-16
DrushDriver::userCreate() doesn't store user id


xaqrox:
Right now it appears to be impossible to create tests that involve creating nodes with specific authorship using the Drush driver. This is because there the user id isn't stored anywhere, and there is no way to map usernames to uids. I found a basically functional solution posted a couple years back at https://www.drupal.org/project/drupalextension/issues/1858858. I'll start a pull request with that fix. It doesn't let you use `author` in the table syntax, but it does let code in custom contexts find the uid, which is what I need right now.

xaqrox:
Added d1cdea2 which now *does* let you do things like
```gherkin
Feature:
  In order to test things
  As a developer
  I need to be able to create content with specific authorship using behat

  @api
  Scenario: Create node with specific authorship
    Given users:
      | name        | mail            |
      | Joe Content | joe@example.com |
    And "page" content:
      | title      | author      |
      | Joe's page | Joe Content |
    And I am logged in as a user with the "administrator" role
    When I visit "admin/content"
    Then I should see "Joe Content" in the "Joe's page" row

jhedstrom:
Since that pr was already merged, without the above commit, can you open a new PR?

PR #184, opened 2018-04-16
Extract and store uid in DrushDriver::userCreate()

xaqrox:
For #183. Credit to @bitamar via https://www.drupal.org/project/drupalextension/issues/1858858


jhedstrom:
This looks good. Thanks for picking up that old issue!

PR #185, opened 2018-04-30
Support deleting any entity, not just content entities.

davereid:
There's no need to special case content entities, this would allow us to delete config entities too.


jhedstrom:
That makes a ton of sense. Thanks!

PR #190, opened 2018-05-27
Drush entity support

arosboro:
This feature extends the Drush driver to support 'entity-create' and 'entity-delete' with the [drush behat endpoint](https://github.com/drush-ops/behat-drush-endpoint/pull/12).

It requires [PR #300](https://github.com/jhedstrom/drupalextension/pull/300/) to be applied to drupal/drupal-extension


jhedstrom:
Thanks! Now the `composer.json` in https://github.com/jhedstrom/drupalextension/pull/300/ can be temporarily updated to point at this commit so we can see if it passes.

PR #191, opened 2018-05-31
Adds field handler for address fields.

greensomes:
Hello. I was recently writing Behat tests for content that includes an Address field (https://www.drupal.org/project/addressfield). Since there is no handler for this complex field type, I decided to spend some time and create one. I simply forked the repo and added the AddressHandler.php class after I did some thorough testing and inline documentation. Please let me know if there are any documentation files that I need to update or automated tests that are required to be created.

Best,
Jeff Schwartz


jhedstrom:
Thanks @greensomes this looks great!

PR #193, opened 2018-07-03
Fixing the ListTextHandler to allow a key to also be 0

upchuk:
Hey there @jhedstrom,

With the current implementation, if the key of the option is 0, the `elseif` will evaluate to FALSE. 

This fix corrects the issue.


nagyad:
Reviewed and tested.

jhedstrom:
Thanks!

Issue #195, opened 2018-08-27
Avoid accidental overwrites of original configuration


jonathanjfshaw:
From #182:

```
+    $config = \Drupal::configFactory()->getEditable('system.mail');
+    $data = $config->getRawData();
+
+    // Save the original values for restoration after.
+    $this->originalConfiguration['system.mail'] = $data;
```

If startCollectingMail is called twice before stopCollectingMail is called (and I think this might happen in the parrallel behat extension PR), then test_mail_collector will end up being stored in $originalConfiguration and persisted.

Therefore we should not overwrite values in originalConfiguration.

PR #196, opened 2018-08-27
Add a storeOriginalConfiguration method

jonathanjfshaw:
Adds a helper method to avoid overwriting original configuration, that stores a value to the original configuration store only if no such value has previously been stored.

### DIFF
--- a/src/Drupal/Driver/Cores/Drupal8.php
+++ b/src/Drupal/Driver/Cores/Drupal8.php
@@ -615,20 +615,17 @@ class Drupal8 extends AbstractCore {
   /**
    * Store the original value for a piece of configuration.
    *
+   * If an original value has previously been stored, it is not updated.
+   *
    * @param string $name
    *   The name of the configuration.
    * @param mixed $value
    *   The original value of the configuration.
-   *
-   * @return bool
-   *   Whether or not $value is now stored as the original configuration value.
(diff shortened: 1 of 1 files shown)


jonathanjfshaw:
Fixes #195, from #182.

src/Drupal/Driver/Cores/Drupal8.php
```diff
@@ -612,4 +612,23 @@ protected function stopCollectingMailSystemMail() {
...
+   *
+   * @return bool
+   *   Whether or not $value is now stored as the original configuration value.
+   */
+  protected function storeOriginalConfiguration($name, $value) {
+    if (!isset($this->originalConfiguration[$name])) {
+      $this->originalConfiguration[$name] = $value;
+      return TRUE;
+    }
+    return ($this->originalConfiguration[$name] === $value);
```
jhedstrom: Is there a reason to have a return value at all here? It doesn't appear to be used in this change...

jonathanjfshaw:
There's no good reason; I was just a bit unhappy about silently ignoring
what the calling code had requested. Probably I was just trying to be too
clever. Shall I remove it?

jonathanjfshaw:
Done

jhedstrom:
Thanks!

Issue #194, opened 2018-08-17
DrupalDriver core needs a configGetOriginal() method


upchuk:
We currently have a `configGet()` method but we also need a `configGetOriginal()` so that we can use in the extension in the `ConfigContext`. This is so that it backs up the correct config value and not the potentially overridden one.

PR will follow.

jhedstrom:
Closing this out. I just restarted the build on the DrupalExtension PR, so we can follow-up there if it needs more work.

PR #197, opened 2018-08-27
Added a method configGetOriginal which will return the original config data.

aritomelo:
**Motivation**:
- DrupalDriver configGet() method which uses the Drupal "get" method to retrieve the configuration. The problem is that the latter takes override into account which means that when the Behat extension restores the config backups, it will save the override instead of the original config value.

By providing this method, when the Behat extension restores the config it will take the original data from configuration (eg.:settings.php).

### DIFF
--- a/src/Drupal/Driver/Cores/CoreInterface.php
+++ b/src/Drupal/Driver/Cores/CoreInterface.php
@@ -210,6 +210,19 @@ interface CoreInterface {
   public function configGet($name, $key = '');
 
   /**
+   * Returns the original configuration item.
+   *
+   * @param string $name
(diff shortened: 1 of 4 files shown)


upchuk:
@jhedstrom This one fixes https://github.com/jhedstrom/DrupalDriver/issues/194. And it will be used by the extension, see issue https://github.com/jhedstrom/drupalextension/issues/501. A PR will follow also for that.

aritomelo:
@jhedstrom Hi mate the second PR can be found here: https://github.com/jhedstrom/drupalextension/pull/503

jhedstrom:
Cool, that makes sense. Thanks!

PR #198, opened 2018-09-21
Use cache:rebuild instead of cache-clear all with Drush 9.

greg-1-anderson:
See https://github.com/jhedstrom/drupalextension/pull/506


jhedstrom:
Thanks!

Issue #200, opened 2018-11-01
'Date only' date type field is not being created while running behat tests


amitsharmadoj:
After  upgrading drupal core from 8.3.9 to 8.6.1, behat from 3.1.0 to  3.3.1 and drupal-driver from 1.2.x-dev to 2.0.x-dev, behat tests are failing for dates of 'Date only' date type. D8 has two date types 'Date and time' and 'Date only' for 'Date' field. There  is no issue with 'Date and time' date type. It looks drupal-driver is not able to create dates of 'Date only' date type. I have tried different date formats but none is working. 
I can create pages with 'Date only' dates and they appear correctly on  the page in the desired format, it is just the behat test which is failing to generate it on the page. It looks DatetimeHandler class in not handling 'Date only' date type.

My **test.feature** looks like (field_publication_date is 'Date only' date) 
@api 
Feature: I can see publication data on publication page. 
Background: 
Given I am an anonymous user 
And "publication" content:
| title          | moderation_state | field_publication_author | field_publication_date |
| test-title   | published              | test-author                     | 2016-09-01                 |

Scenario: I can see the correct publication data 
Given I am on "/test-title" 
And   I should see the text "test-author" 
And   I should see the text "September 2016" 

**Error:** 
001 Scenario: I can see publication data on publication page # features/publication.feature:35 And I should see the text "September 2016"  # features/publication.feature:44 The text "September 2016" was not found anywhere in the text of the current page. (Behat\Mink\Exception\ResponseTextException) 

**Format tried:** 
2016-09-01, 2016/09/01, 01/09/2016, 01-09-2016, 2016-09-01T05:00:00, 2016-09-01 12:12:12, 1 minutes ago etc.
As I said it doesn't looks like format issue. Behat is not able to create date of 'Date only' type. I have also tried putting break just before date test statement to see  if date is being created in different format other than expected but  date field along with its label is not at all present on the page. 

**Versions used:** 
```
behat/behat:    v3.3.1 
behat/gherkin:    v4.5.1 
behat/mink:    v1.7.1 
drupal/drupal-driver:   2.0.x-dev
drupal/core:    8.6.1
```
--

amitsharmadoj:
Patch created and it resolved my issue
https://patch-diff.githubusercontent.com/raw/jhedstrom/DrupalDriver/pull/201.patch

bkosborne:
[...]

PR #201, opened 2018-11-01
Issue 200 fix issue with creation of date only fields

amitsharmadoj:
Fix issue 200 (https://github.com/jhedstrom/DrupalDriver/issues/200), an issue with the creation of date only fields

### DIFF
--- a/src/Drupal/Driver/Fields/Drupal8/DatetimeHandler.php
+++ b/src/Drupal/Driver/Fields/Drupal8/DatetimeHandler.php
@@ -28,7 +28,7 @@ class DatetimeHandler extends AbstractHandler {
         // field value by the step author, assume the default timezone of
         // the Drupal install, and therefore transform it into UTC for storage.
         if (DateTime::createFromFormat('Y-m-d', $value) !== FALSE) {
-          // Handle 'Date only' date type
+          // Handle 'Date only' date type.
           $date = new DateTime($value);
           $formattedDate = $date->format('Y-m-d');
         }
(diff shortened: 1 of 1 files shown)


src/Drupal/Driver/Fields/Drupal8/DatetimeHandler.php
```diff
@@ -27,9 +27,17 @@ public function expand($values) {
         // uses UTC for internal storage. If no timezone is specified in a date
         // field value by the step author, assume the default timezone of
         // the Drupal install, and therefore transform it into UTC for storage.
-        $date = new DateTime($value, $siteTimezone);
-        $date->setTimezone($storageTimezone);
-        $values[$key] = $date->format('Y-m-d\TH:i:s');
+        if (DateTime::createFromFormat('Y-m-d', $value) !== FALSE) {
```
jhedstrom: Would this ever return `FALSE`?

jhedstrom:
The reason the old code uses that format is that even for date-only fields, the actual storage has the entire date and time string.

amitsharmadoj:
Yes this is what my understanding is. Ideally date-only fields should have been created with and without the timestamp but not happening with timestamp for me. Perhaps there is a change in Drupal core 8.6.x the way it handles the Date.

MPParsley:
Running into the same issue, I have a date only field but I can set the value:

```
    When I am viewing a "page" content:
      | langcode    | nl                                   |
      | title       | Title                     |
      | field_date | 2020-07-14           |
```

field_date remains empty

MPParsley (approved):

AlexSkrypnyk:
Hey @amitsharmadoj - thank you for this contribution! Date-only field handling has been fixed in #309 which was recently merged. I'm closing this as superseded. Appreciate you reporting and working on this!

PR #211, opened 2019-10-08
Fixes #210: Don't throw exception when deleting non-existant role

### LINKED ISSUE #210: Don't throw an exception when attempting to delete role that doesn't exist
In RawDrupalContext, there is an AfterScenario that deletes roles that were programmatically created by other step definitions that were executed:

```
  /**
   * Remove any created roles.
   *
   * @AfterScenario
   */
  public function cleanRoles() {
    // Remove any roles that were created.
    foreach ($this->roles as $rid) {
      $this->getDriver()->roleDelete($rid);
    }
    $this->roles = array();
  }
```

For the Drupal8 core, this code gets run:

```
  /**
   * {@inheritdoc}
   */
  public function roleDelete($role_name) {
    $role = user_role_load($role_name);

    if (!$role) {
      throw new \RuntimeException(sprintf('No role "%s" exists.', $role_name));
    }

    $role->delete();
  }
```

The problem is that the role may no longer exist, since actions taken during the step may have deleted the role. This exception causes the test to fail even though everything is fine.

I noticed that the other "delete" methods on the Drupal8 core first check that the entity being deleted exists before deleting it. I think we should do the same here.

bkosborne:
Fixes #210

### DIFF
--- a/src/Drupal/Driver/Cores/Drupal8.php
+++ b/src/Drupal/Driver/Cores/Drupal8.php
@@ -13,6 +13,7 @@ use Drupal\node\NodeInterface;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\taxonomy\Entity\Term;
 use Drupal\taxonomy\TermInterface;
+use Drupal\user\RoleInterface;
 use Symfony\Component\HttpFoundation\Request;
 
 /**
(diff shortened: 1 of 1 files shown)


pfrenssen (approved):

AlexSkrypnyk:
Hey @bkosborne - thanks for this contribution! I've opened #319 which builds on your fix. Your contribution is attributed there. I'll close this PR once the new one is merged. Appreciate the work!

PR #214, opened 2019-12-05
Fix D8 deprecations.

vever001:
Hello,
Should we fix all those deprecations from D8?
If so here it is.


jhedstrom:
Awesome! Thank you!

PR #222, opened 2020-02-17
Allow Symfony 4 as dependency

Berdir:
Requiring ~4.4 as that's what Drupal 9 requires anyway.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,8 +13,8 @@
   ],
   "require": {
     "php": ">=5.6",
-    "symfony/process": "~2.5|~3.0",
-    "symfony/dependency-injection": "~2.6|~3.0",
+    "symfony/process": "~2.5|~3.0|~4.4",
+    "symfony/dependency-injection": "~2.6|~3.0|~4.4",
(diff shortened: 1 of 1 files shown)


Berdir:
PHP 7.4 fails seem unrelated to this, phpunit and other dependencies will need to be updated to supported versions to fix that.

jhedstrom:
Hmm, what should we do regarding the php 7.4 failure? I'm assuming this is still being sorted out in Drupal itself?

Berdir:
Latest drupal 8 dev should be 7.4 compatible, that doesn't seem to be the problem here but your own dependencies as far as I see. 

You still require PHPUnit 4, Drupal 8 uses 6/7 while Drupal 9 uses Phpunit 8. And some other dependencies that might need to be updated too.

I'm not sure why 7.4 is even currently tested, this can't work right now in HEAD either?

Berdir:
Having a look at how much work it is to update those dependencies and if things break, but to me it would make more sense to do that in a separate PR :)

PR #223, opened 2020-02-19
Dev dependencies

Berdir:
Updating necessary dev dependencies to try and get it working on more recent php versions.

only have php 7.3 locally, might need to allow 8 too which might require more changes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
   },
   "require-dev": {
     "drupal/coder": "~8.3.0",
-    "phpspec/phpspec": "~2.0 || ~4.0",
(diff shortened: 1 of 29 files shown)


Berdir:
the drupal/coder update might not be necessary, but I think the coding standard improvements make sense, although likely to conflict with some other PR's.

Berdir:
Ok, this passes despite a deprecation warning in jakub-onderka/php-parallel-lint which doesn't seem to be really maintained anymore. Might be possible to drop this, speed doesn't really seem like an issue here?

jhedstrom:
> Might be possible to drop this, speed doesn't really seem like an issue here?
Sure, we can do that in a follow-up, but I merged this for now.

PR #224, opened 2020-03-14
Support Drupal 9

Berdir:
I'm not 100% sure, but considering how Drupal 9+ is expected to behave, this seems like a simple option.

Able to do some first tests against Drupal 9 with this.


jhedstrom:
Thanks!

PR #227, opened 2020-06-22
Also clear static cache of EntityAccessControlHandler

### LINKED ISSUE #226: Also clear static cache of EntityAccessControlHandler
We are having an issue with the entity access cache containing stale items in Behat scenarios. The problem occurs in the following circumstances (all the following steps are executed in a single Behat scenario, i.e. in a single Drupal request)

1. Use an entity type which implements `EntityPublishedInterface` does not implement `RevisionableInterface`.
1. Create an unpublished entity.
1. View the entity with a user which does not have permission to view unpublished entities.
1. Publish the entity.
1. View the entity again with the same user.

Expected result: the user should see the entity.
Actual result: the entity is invisible.

This is caused by the static cache of `EntityAccessControlHandler`. The access status of entities are cached using a cache ID that is comprised of the UUID or `{entity_type_id}:{entity_id}`. In case of a revisionable entity type the revision ID is also appended to the cache ID.

In the following circumstances this works fine:

* When using Drupal through the UI - the static cache only lives as long as a single request, and on subsequent requests it is rebuilt. It is unlikely to hit stale caches in these circumstances.
* When using revisionable entities the cache ID changes as soon as the entity is published, and the stale cache entry is never hit.

However in a case of using a non-revisioned entity inside a long running process such as a Behat test these stale caches become a problem.

Suggested solution: clear the EntityAccessControlHandler` static cache in `\Drupal\Driver\Cores\Drupal8::clearStaticCaches()` so that developers can decide to clear the caches in a related step definition of hook.

pfrenssen:
Fixes #226

### DIFF
--- a/src/Drupal/Driver/Cores/Drupal8.php
+++ b/src/Drupal/Driver/Cores/Drupal8.php
@@ -464,6 +464,9 @@ class Drupal8 extends AbstractCore implements CoreAuthenticationInterface {
   public function clearStaticCaches() {
     drupal_static_reset();
     \Drupal::service('cache_tags.invalidator')->resetChecksums();
+    foreach (\Drupal::entityTypeManager()->getDefinitions() as $definition) {
+      \Drupal::entityTypeManager()->getAccessControlHandler($definition->id())->resetCache();
+    }
   }
 
   /**


pfrenssen:
Travis CI didn't report back the test result, here it is: https://travis-ci.org/github/jhedstrom/DrupalDriver/builds/700827873

AlexSkrypnyk:
Hey @pfrenssen - thanks for this contribution! I've opened #318 which builds on your fix. Your contribution is attributed there. I'll close this PR once the new one is merged. Appreciate the work!

Issue #230, opened 2020-11-10
"I am logged in as a user with the..." steps fail if user name is changed when created


rsanzante:
**Brief**
DrupalDriver assumes user name is not modified when a user is created. Thus, if you override the name by any means (in my case, in the entity insert hook), Behat is unable to login the given user because the username is not the one Behat expects.

**Details**

When DrupalDriver creates a user it uses, for example in Drupal 8,  Drupal\Driver\Cores\Drupal8::userCreate.

This function creates the user base on data received from the parameter $user. However, DrupalDriver doesn't take into account any modifications, so if username has changed during entity insertion DrupalDriver is not aware of this.

Is interesting that DrupalDriver does update the user id. 

**Solution**

Not sure. Maybe just update the username as the id is updated. This is a solution for the name field. Any other user properties are not updated. However, the data DrupalDriver keeps (or more precisely, DrupalExtension) is used for login the user, so maybe it is ok if other fields are not updated.

rsanzante:
Added a PR with a simple PoF for D8. It just updates the user name.

pfrenssen:
Changing the username on creation is not default Drupal behavior, so this is not something we support out of the box.

I would invite you to have a look at the `@AfterUserCreate` hook. If you implement this hook you can do whatever changes are needed in scope of your test coverage to match your custom code.

Possibly something like this could fit your use case? (not tested)
```
  /**
   * Update the user name after a new user is created, to match our custom functionality.
   *
   * @Berdir fterUserCreate
   */
  public static function updateUsernameAfterUserCreate(EntityScope $scope) {
    $user = $scope->getEntity();
    $entity = User::load($user->uid);
    $user->name = $entity->getUsername();
  }
```

rsanzante:
I didn't know about that hook (or I forgot about it 😅).

Your code did the trick without any modifications (well, just adding the `use Drupal\user\Entity\User;`).

Thank you very much, I really appreciate your help.

PR #231, opened 2020-11-11
Update username in Behat data after creating it in Drupal.

rsanzante:
See https://github.com/jhedstrom/DrupalDriver/issues/230.


rsanzante:
This is not complete because it only addresses the issue on D8 core. If solution si accepted it can be easily extended to D7.

pfrenssen:
I think it is probably not needed to change this since it should be handled on the calling side. See my comment https://github.com/jhedstrom/DrupalDriver/issues/230#issuecomment-725379086 that proposes to look into how Behat Drupal Extension deals with possible altering of the user object using the `@AfterUserCreate` hook.

rsanzante:
Thanks for your comment, I agree calling side os a good place to handle this.

PR #235, opened 2021-05-10
Fix date only field not working.

drishu:
The patches on issue https://github.com/jhedstrom/DrupalDriver/issues/200 where either conflicting or just not working, this resolves it for latest master.

### DIFF
--- a/src/Drupal/Driver/Fields/Drupal8/DatetimeHandler.php
+++ b/src/Drupal/Driver/Fields/Drupal8/DatetimeHandler.php
@@ -21,14 +21,14 @@ class DatetimeHandler extends AbstractHandler {
         // Get time, convert to ISO 8601 date in GMT/UTC, remove TZ offset.
         $values[$key] = substr(gmdate('c', strtotime($relative)), 0, 19);
       }
-      else {
+      elseif ($this->fieldInfo->getSetting('datetime_type') === 'datetime') {
         // A Drupal install has a default site timezone, but nonetheless
         // uses UTC for internal storage. If no timezone is specified in a date
         // field value by the step author, assume the default timezone of
         // the Drupal install, and therefore transform it into UTC for storage.
         $date = new \DateTime($value, $siteTimezone);
         $date->setTimezone($storageTimezone);
-        $values[$key] = $date->format('Y-m-d\TH:i:s');
+        $values[$key] = $date->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT);
       }
     }
     return $values;


src/Drupal/Driver/Fields/Drupal8/DatetimeHandler.php
```diff
@@ -26,9 +26,16 @@ public function expand($values) {
         // uses UTC for internal storage. If no timezone is specified in a date
         // field value by the step author, assume the default timezone of
         // the Drupal install, and therefore transform it into UTC for storage.
-        $date = new \DateTime($value, $siteTimezone);
-        $date->setTimezone($storageTimezone);
-        $values[$key] = $date->format('Y-m-d\TH:i:s');
+        if ($this->fieldInfo->getSetting('datetime_type') === 'date') {
+          $date = new \DateTime($value);
```
ademarco, line 30: Shouldn't we keep applying the storage timezone, even on the date type? The only thing we need to change is the storage format, which depends on the field type, but we should keep the initial conversion into UTC, that's necessary, as explained in the comment above.

drishu: From my tests the answer is no, but I need to still debug to see why

drishu: I didn't exactly find mathematical proof here, but, since the date has no time it should not have any timezone conversion, the comment you mention was written for the case time is set but not timezone, not our case. The date will be saved as is, and the storage timezone is already taken into consideration here https://git.drupalcode.org/project/drupal/-/blob/9.2.x/core/modules/datetime/src/Plugin/Field/FieldFormatter/DateTimeFormatterBase.php#L168

ademarco: Why would we need a datetime object, if we have no time? If we are anyway expecting the step author to provide the date in the DateTimeItemInterface::DATE_STORAGE_FORMAT, can't we just apply the timezone conversion for datetime fields, while simply pass along fields having `datetime_type: date`?

ademarco: I mean, wouldn't this work?

```php
  /**
   * {@inheritdoc}
   */
  public function expand($values) {
    $siteTimezone = new \DateTimeZone(\Drupal::config('system.date')->get('timezone.default'));
    $storageTimezone = new \DateTimeZone(DateTimeItemInterface::STORAGE_TIMEZONE);
    foreach ($values as $key => $value) {
      if (strpos($value, "relative:") !== FALSE) {
        $relative = trim(str_replace('relative:', '', $value));
        // Get time, convert to ISO 8601 date in GMT/UTC, remove TZ offset.
        $values[$key] = substr(gmdate('c', strtotime($relative)), 0, 19);
      }
      elseif ($this->fieldInfo->getSetting('datetime_type') !== 'date') {
        // A Drupal install has a default site timezone, but nonetheless
        // uses UTC for internal storage. If no timezone is specified in a date
        // field value by the step author, assume the default timezone of
        // the Drupal install, and therefore transform it into UTC for storage.
        $date = new \DateTime($value, $siteTimezone);
        $date->setTimezone($storageTimezone);
        $format = DateTimeItemInterface::DATETIME_STORAGE_FORMAT;
        $values[$key] = $date->format($format);
      }
    }
    return $values;
  }
```

BramDriesen:
Tested and reviewed <3

cboyden-ucb:
Likewise - this PR fixes the issue #200 where the Given Content... step fails to fill in the datetime field if that field is date-only.

VladimirAus:
Tested and ready to be merged. +1

vever001:
+1 on this, works great, thanks!

AlexSkrypnyk:
Hey @drishu - thank you for this contribution! Date-only field handling has been fixed in #309 which was recently merged. I'm closing this as superseded. Appreciate the work!

PR #241, opened 2022-01-18
Allow Entity Reference Field Handler to Refernce by ID

kerasai:
I'm using a fair amount of custom steps to handle complicated data setup within the tests, which I track and reference via https://packagist.org/packages/kerasai/behat-tokenizer.

This PR changes the D8+ entity reference field handler to find the referenced entity by label (existing functionality) or ID.


kerasai:
Note, this is a similar fix to #94 but retains backwards compatibility.

Also related to #117 and #133.

jhedstrom:
Thank you!

PR #243, opened 2022-03-07
Expose ::isBaseField()

### LINKED ISSUE #242: Expose ::isBaseField() on the public interfaces
In #114 we introduced a new public `Drupal8::isBaseField()` method, but this was not exposed on the interfaces since this was Drupal 8 only.

In the meantime Drupal 8+ has become dominant and it would be interesting to expose this method so we can check for base fields.

It would be also interesting to discuss whether we should deal with the distinction between base fields and configurable fields at all. In Drupal 8+ these are handled transparently by the Field API and it would probably make more sense to remove `::isBaseField()` and just have `::isField()` return TRUE for both base fields and configurable fields. But unfortunately this would break backwards compatibility.

pfrenssen:
Fixes #242

### DIFF
--- a/src/Drupal/Driver/BaseDriver.php
+++ b/src/Drupal/Driver/BaseDriver.php
@@ -136,6 +136,13 @@ abstract class BaseDriver implements DriverInterface {
   /**
    * {@inheritdoc}
    */
+  public function isBaseField($entity_type, $field_name) {
+    return FALSE;
+  }
+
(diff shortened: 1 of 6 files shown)


pfrenssen:
Tests are failing due to stricter coding standards checks, these are fixed in #244.

jhedstrom:
Makes sense!

Issue #249, opened 2022-08-01
Avoid the composer require of core "drupal/core-utility"


dxvargas:
In Drupal 9.4.4 the replace section of ```composer.json``` of ```drupal/core``` has been removed.
See here https://www.drupal.org/node/3295096.

There is a plan to remove core dependencies from composer.json. [Here](https://www.drupal.org/project/drupal/issues/3292380) we can read in the "Remaining tasks" section:
- Search for modules that require core modules in their composer.json. Most probably we want to file patches against those where we remove the require, and move those to the dependencies: section of their info.yml file.

What about here in drupal-driver? It's requiring ```drupal/core-utility```
https://github.com/jhedstrom/DrupalDriver/blob/v2.2.0/composer.json#L18

When we run ```composer require drupal/drupal-driver```, we end up with duplicated ```drupal/core-utility``` in the vendor folder! It overrides the core package in ```web/core/lib/Drupal/Component/Utility``` and may even have a different version.

Should we avoid the composer require of "drupal/core-utility"?

fgm:
See also https://www.drupal.org/project/drupal/issues/3116405#comment-14635486

AlexSkrypnyk:
This will be assessed as part of the v3.x effort (see #312).

AlexSkrypnyk:
Thanks for raising this back in 2022. I checked the current state against Drupal 11.3.7 and the concern no longer manifests in practice: drupal/core still ships a replaces entry for drupal/core-utility self.version, so composer sees this library's requirement as satisfied by drupal/core itself. No separate core-utility package is installed under vendor, and Random loads from Drupal core's own lib/Drupal/Component/Utility/Random.php. Whatever the 9.4.4 replaces change touched, it did not include core-utility or it was reverted upstream. Closing as obsolete. If a future Drupal release drops the replaces entry we would see duplication immediately in CI and can reopen with a reproducer at that point.

PR #250, opened 2022-10-06
Added Rector and Drupal Check and fixed Drupal 10 compatibility.

AlexSkrypnyk:
- Added Rector
- Added Drupal Check
- Fixed code that was not compatible with Drupal 10
- Fixed compatibility with `Symfony\Process` in Drush Driver (copied from https://github.com/jhedstrom/DrupalDriver/pull/251 - sorry, had to keep it in this branch for downstream projects)

Please note that Rector and Drupal check are only scanning Drupal 8 driver.

Build__767_-_jhedstrom_DrupalDriver_-_Travis_CI


.travis.yml
```diff
@@ -1,8 +1,6 @@
 language: php
 
 php:
-  - 7.4
-  - 8.0
```
pfrenssen, line 5: Any reason to stop testing on PHP 8.0? It is still supported until October 2023.

I think for as long as we declare in composer.json to be compatible with PHP 7.4 we should also keep testing on 7.4. Probably we shouldn't drop PHP 7.4 support unless we plan to release a new major version.

I suppose the `./vendor/bin/drupal-check` test only works on PHP 8.1, but we can still run the remainder of the suite on older PHP versions.

composer.json
```diff
@@ -15,25 +15,34 @@
...
   },
   "scripts": {
     "test": [
       "composer validate --no-interaction",
       "parallel-lint src spec tests",
       "phpunit",
       "phpspec run -f pretty --no-interaction",
-      "phpcs --standard=./phpcs-ruleset.xml ."
-
+      "phpcs --standard=./phpcs-ruleset.xml --ignore=drupal .",
```
pfrenssen, line 41: Let's not add the `--ignore` option inline. Let's instead update the ignored paths in `phpcs-ruleset.xml`:

```
<exclude-pattern>./drupal/*</exclude-pattern>
```

src/Drupal/Driver/Cores/Drupal8.php
```diff
@@ -385,7 +392,7 @@ public function getExtensionPathList() {
    * @param array $base_fields
    *   Base fields to be expanded in addition to user defined fields.
    */
-  public function expandEntityBaseFields($entity_type, \stdClass $entity, array $base_fields) {
+  public function expandEntityBaseFields($entity_type, $entity, array $base_fields) {
```
pfrenssen, line 395: Why remove the `\stdClass` type hint? This is here to clarify that this is not a real entity but rather a "mocked" value object. It is literally a `\stdClass` object so the type hint is correct. Am I missing something?

src/Drupal/Driver/Fields/Drupal8/ImageHandler.php
```diff
@@ -17,9 +17,7 @@ public function expand($values) {
     }
 
     /** @var \Drupal\file\FileInterface $file */
-    $file = file_save_data(
-      $data,
-      'public://' . uniqid() . '.jpg');
+    $file = \Drupal::service('file.repository')->writeData($data, 'public://' . uniqid() . '.jpg');
```
pfrenssen, line 20: This has been introduced in Drupal 9.3, so this breaks compatibility with Drupal 8 and Drupal 9 up to version 9.2.x

I am not opposed to this, but if this gets in we have to clearly stop supporting Drupal 9.2 and below. We need to add this to `composer.json`:

```
"conflict": {
    "drupal/core": ">=8.0 <9.3"
},
```

This will ensure that any projects stuck on an old version of Drupal for any reason will not accidentally break if they update DrupalDriver. They will remain on the last supported version.

src/Drupal/Driver/Cores/Drupal8.php
```diff
@@ -4,6 +4,7 @@
 
 use Drupal\Core\DrupalKernel;
 use Drupal\Core\Field\BaseFieldDefinition;
+use Drupal\Core\Routing\RouteObjectInterface;
```
pfrenssen, line 7: This has been introduced in Drupal 9.1, so adding this will break compatibility with Drupal 8.x and 9.0. I think it's OK to introduce this but we need to indicate this incompatibility in our `composer.json` file, using the `conflict` directive I mentioned in another comment.

pfrenssen (changes requested):

.travis.yml
AlexSkrypnyk: Drupal 10's requirement is to use 8.1. The CI may fail for 8.0 on D10

src/Drupal/Driver/Cores/Drupal8.php
AlexSkrypnyk: This, as well as other changes, is a result of Rector's and drupal-check report. I don't remember what static analysis rule this was violating specifically, but it would not pass those tools.

We can revert and add an ignore comment for those tools to ignore this. 

Please advise

AlexSkrypnyk:
@pfrenssen 
Thank you for reviewing this. Agree with all comments. I will address them shortly. Thanks

src/Drupal/Driver/Fields/Drupal8/ImageHandler.php
AlexSkrypnyk: added

.travis.yml
AlexSkrypnyk: PHP8 did not work

```
Your requirements could not be resolved to an installable set of packages.
  Problem 1
    - drupal/core-recommended[10.0.0-beta1, ..., 10.0.0-beta2] require symfony/process ~v6.1.3 -> satisfiable by symfony/process[v6.1.3].
    - symfony/process[v6.1.0-BETA1, ..., v6.1.3] require php >=8.1 -> your php version (8.0.23) does not satisfy that requirement.
    - Root composer.json requires drupal/core-recommended ^10@beta -> satisfiable by drupal/core-recommended[10.0.0-beta1, 10.0.0-beta2].
The command "composer install" failed and exited with 2 during .
```

https://app.travis-ci.com/github/jhedstrom/DrupalDriver/jobs/585619255

I've reverted my change so that we are only testing on PHP 8.1

If there is a different way to test with PHP 8.0 - please let me know and I will update this PR.

AlexSkrypnyk:
@pfrenssen 
I have addressed all the comments. Could you please review again. Thank you

jhedstrom:
re: testing on previous PHP versions

I think it is these lines that are causing those to not work:

```
    "drupal/core-composer-scaffold": "^10@beta",
    "drupal/core-recommended": "^10@beta",
```

They can probably either be removed, or switched to allow the older versions of Drupal as in the `require` section:

```
    "drupal/core-utility": "^8.4 || ^9 || ^10.@beta"
```

AlexSkrypnyk:
@jhedstrom 
could you please advise what do they have to be switched to?

jhedstrom:
I think they just need to allow 8.4 and 9 like we do for the `core-utility`:

```
    "drupal/core-composer-scaffold": "^8.4 || ^9 || ^10@beta",
    "drupal/core-recommended": "^8.4 || ^9 || ^10@beta",
```

(we could probably drop 8.x for now too, and then drop 9 when it is eol.)

AlexSkrypnyk:
thank you @jhedstrom 
I've updated the configs, but ran out of OSS credits on Travis CI (can we switch to Circle or GH actions?). Not sure how to test this without those credits. Probably have to wait till Nov 1.

jhedstrom:
@AlexSkrypnyk I've opened #254 to track that. It should be a relatively small effort.

Also, there's https://github.com/jhedstrom/DrupalDriver/issues/249 that might be relevant to the version changes happening here for core libraries.

src/Drupal/Driver/Cores/Drupal8.php
claudiu-cristea: @AlexSkrypnyk @pfrenssen I've fixed this in https://github.com/jhedstrom/DrupalDriver/pull/253/files by hardcoding the constant value, so we keep compatibility with Drupal 8/9

AlexSkrypnyk:
@claudiu-cristea @jhedstrom 
I've updated this PR with the latest changes from `master` and updated CI to include D10. Could you please review. Thank you

jhedstrom:
Thanks!

Kingdutch:
@jhedstrom This PR introduced a conflict with older Drupal versions which seems like something that shouldn't happen in a `patch` release (but should be a major version release instead). This prevents upgrading in some instances.

See https://github.com/goalgorilla/open_social_dev/pull/51 for the conflict message that this causes.

PR #251, opened 2022-10-12
Fixed compatibility with Symfony\Process v4+ for Drush driver.

AlexSkrypnyk:
DrushDriver uses `Symfony\Process` to issue CLI commands. 

`Process` expects the first element to be an array. Since we are passing the Drush command with arguments as a string, it would be better to use `Process::fromShellCommandline()` instead of `new Process()` to make sure compatibility with different versions of `Symfony\Process`.

Without this change, drush steps are failing with Symfony v6


jhedstrom:
I've merged #261. Thanks for this!

PR #252, opened 2022-10-13
Allow reference field properties to be preserved after expanding the values.

idimopoulos:
### Problem description
I am having a file field which has the display field enabled. That means that when we create a new entity, we want to also be able to set the 'display' property of the field.
Now, we are using behat but we have arranged our code to reach the proper structure of the field in the `\Drupal\DrupalExtension\Context\RawDrupalContext::nodeCreate` where the values reach as
```
print_r($node->field_attachment);
Array
(
    [0] => Array
        (
            [target_id] => empty.rdf
            [display] => 1
        )

    [1] => empty_pdf.pdf
    [2] => invalid_adms.rdf
    [3] => test.zip
    [4] => text.pdf
)
```
One of the values is a multi column value and the rest are simple label references. Of course we take care of creating the files before hand.
However, when reaching `\Drupal\Driver\Fields\Drupal8\EntityReferenceHandler::expand` the entity query fails with the following error:
```
SQLSTATE[HY093]: Invalid parameter number: number of bound variables does not match number of tokens: SELECT "base_table"."fid" AS "fid", "base_table"."fid" AS "base_table_fid"
FROM
"file_managed" "base_table"
INNER JOIN "file_managed" "file_managed" ON "file_managed"."fid" = "base_table"."fid"
WHERE "file_managed"."filename" = :db_condition_placeholder_0:db_condition_placeholder_1; Array
(
    [:db_condition_placeholder_0] => text.pdf
    [:db_condition_placeholder_1] => 1
)
```
because it simply passes the value in the query.

### Proposal
From what I see from the PHPDocBlocks, the `$values` array is supposed to be a field value array and the return is supposed to be a field value array as supposed to be retrieved by the field storage.
My assumption as such, is that we need to preserve the columns when querying the values and returning them. Thus, I propose to identify the `target_id` property - or any property used for the reference, and if present, extract the label for the entity instead of blindly passing the `$value` in the query.

Please, note that this means that we maintain the philosophy of the current functionality, that since the user passes a label as a value to get the entity, if they would like to use a more complex value, then the label would be passed in the reference property key. For example, if we have a reference to a node, the complex array would be
```
Array
(
    'target_id' => 'My node title'
)
```
and not
```
Array
(
   'title' => 'My node title'
)
```
At least that is how I see it.

### Remaining tasks
Tests?

### DIFF
--- a/src/Drupal/Driver/Fields/Drupal8/EntityReferenceHandler.php
+++ b/src/Drupal/Driver/Fields/Drupal8/EntityReferenceHandler.php
@@ -11,11 +11,8 @@ class EntityReferenceHandler extends AbstractHandler {
    * {@inheritdoc}
    */
   public function expand($values) {
-    $return = [];
     $entity_type_id = $this->fieldInfo->getSetting('target_type');
     $entity_definition = \Drupal::entityTypeManager()->getDefinition($entity_type_id);
-
-    $id_key = $entity_definition->getKey('id');
 
     // Determine label field key.
     if ($entity_type_id !== 'user') {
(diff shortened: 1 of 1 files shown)


idimopoulos:
Testing is a bit trickier because the reference handler uses the container but Drupal is not a dependency in composer and composer requires symfony console which conflicts.
```
 - drupal/core[9.4.0, ..., 9.4.8] require symfony/console ^4.4 -> found symfony/console[v4.4.0, ..., v4.4.47] but the package is fixed to v5.4.14 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.
```

claudiu-cristea:
@jhedstrom Added tests and moved to GitHub Actions. Until GitHub actions are enabled here, you can see tests green in the fork repo https://github.com/idimopoulos/DrupalDriver/actions/runs/3329054956

claudiu-cristea (approved):

claudiu-cristea:
Fixes #254

AlexSkrypnyk:
Hey @idimopoulos - thanks for this contribution! This is a valuable enhancement that preserves secondary properties on entity reference fields. We're planning to tackle this as part of the v3.x effort (see #312) where we'll have kernel tests to properly validate the changes against real Drupal field storage. I'll leave this open as reference for that work. Appreciate the thoughtful implementation!

AlexSkrypnyk:
Hey @idimopoulos - thanks for this contribution! It's been a while since the last update here, and master has moved on significantly (v3.x restructure with capability interfaces and kernel tests), so a clean rebase wasn't viable - but the bug you identified is. I've opened #347 which reimplements the fix against current master: EntityReferenceHandler::expand() now handles associative-array deltas like ['target_id' => 'alice', 'display' => 1] and preserves the extra item columns (display, target_revision_id, etc.) through resolution. Your contribution is attributed at the top of #347, and the PR ships RED to GREEN kernel tests that reproduce the exact InvalidQueryException before the fix and pass after. I will close this PR once #347 is merged. Appreciate you catching this!

AlexSkrypnyk:
Landed via #347. Thanks again, @idimopoulos!

PR #257, opened 2022-11-29
Fix 'id' not found

MPParsley:
The following scenario results in an error `'id' not found (Drupal\Core\Entity\Query\QueryException)` on php 8.1:

```
    Given I am logged in as an "administrator"
    And "my_taxonomy" terms:
      | name | parent | description |
      | 123  |        | Description |
```

Instead of checking isset we should check if not empty
Or there should be a way to pass `NULL` in the NodeTable.


Kingdutch:
Please use https://github.com/jhedstrom/DrupalDriver/pull/260 instead

MPParsley:
Ok thanks for pointing that out!

PR #258, opened 2022-11-30
Document testing locally

xurizaemon:
They say the best way to find out if something is true is to post it on the internet.

Does this correctly document how to run tests locally before submitting a PR?


jhedstrom:
Thank you!

PR #260, opened 2022-12-02
Fix taxonomy term creation

### LINKED ISSUE #259: Drupal 10 compatibility broke taxonomy term creation in patch version
https://github.com/jhedstrom/DrupalDriver/pull/250/commits/8177f645a229f40ba88b4c20daffe1a124c54445 changed how taxonomy term creation worked in 2.2.1 but this broke the functionality.

`parent` may be empty if multiple taxonomies are created in a single `Given` block as pointed out in https://github.com/jhedstrom/DrupalDriver/pull/257

Additionally `id` is not the right base field for the query but it should be `name`.

Finally the entityQuery does not return loaded entities but returns IDs so calling `->id()` on it will break as well.

Kingdutch:
`parent` may be empty if multiple taxonomies are created in a single `Given` block.

Additionally `id` is not the right base field for the query but it should be `name`.

Finally the entityQuery does not return loaded entities but returns IDs so calling `->id()` on it will break as well.

Fixes #259

### DIFF
--- a/src/Drupal/Driver/Cores/Drupal8.php
+++ b/src/Drupal/Driver/Cores/Drupal8.php
@@ -333,16 +333,14 @@ class Drupal8 extends AbstractCore implements CoreAuthenticationInterface {
   public function termCreate(\stdClass $term) {
     $term->vid = $term->vocabulary_machine_name;
 
-    if (isset($term->parent)) {
+    if (!empty($term->parent)) {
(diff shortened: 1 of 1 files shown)


MPParsley (approved):
Looks good 👍

aaron-capellic:
Thank you, this is working for me.

jhedstrom:
Thank you!

PR #261, opened 2023-01-20
Broaden compatibility with symfony/process

jonpugh:
Updates "Fixed compatibility with Symfony\Process v4+ for Drush driver. #251".

I don't know why, I can't explain it. A project has symfony/process 4.4.4, but still, does NOT have the method "fromShellCommandline".

I noticed that another part of symfony/console actually checks for this method to remain compatible, I guess?

Can do this so that the driver works everywhere?

```php
    $process = method_exists(Process::class, 'fromShellCommandline') ? Process::fromShellCommandline($cmd) : new Process($cmd);

```

https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Console/Helper/ProcessHelper.php#L57

### DIFF
--- a/src/Drupal/Driver/DrushDriver.php
+++ b/src/Drupal/Driver/DrushDriver.php
@@ -414,7 +414,8 @@ class DrushDriver extends BaseDriver {
     // Add any global arguments.
     $global = $this->getArguments();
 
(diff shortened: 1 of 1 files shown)


jhedstrom:
Looks good to me, thanks!

PR #267, opened 2023-09-13
Populate link fields only with URI if value is a string

zolhorvath:
Problem / motivation
---

As Drupal developer, I would like to create link fields without specifying any value for the field item's `title` property. Right now this seems to be impossible to me.

About this PR:
---
Still the same [D8 LinkHandler](https://github.com/jhedstrom/DrupalDriver/blob/312c0f7721c54f/src/Drupal/Driver/Fields/Drupal8/LinkHandler.php) issue as in #137, but a little bit different approach.
* Title attribute is fully omitted if field value is a string.
* Extra test coverage.

### DIFF
--- a/src/Drupal/Driver/Fields/Drupal8/LinkHandler.php
+++ b/src/Drupal/Driver/Fields/Drupal8/LinkHandler.php
@@ -16,14 +16,18 @@ class LinkHandler extends AbstractHandler {
       // 'options' is required to be an array, otherwise the utility class
       // Drupal\Core\Utility\UnroutedUrlAssembler::assemble() will complain.
       $options = [];
+      // Instantiate the value as URI-only if $value is a string.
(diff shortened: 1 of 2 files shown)


vever001:
I just hit the same issue, there's probably an overlap with https://github.com/jhedstrom/DrupalDriver/pull/277 but +1 on this, works as expected, thanks @zolhorvath !

AlexSkrypnyk:
Hey @zolhorvath - thanks for this contribution! I've picked up the work and iterated on it. I've opened #313 which builds on your changes, re-implemented against the current codebase (which now supports named keys from #311) and includes updated tests. Credit also given to @ohthehugemanatee whose earlier #137 proposed the same feature. I'll close this PR once the new one is merged. Appreciate the work!

PR #271, opened 2024-03-06
Enhance entity creation with base field expansion and dynamic id key

### LINKED ISSUE #270: Entity creation does not expand base fields & assumes "id" for entity ID key (D8/9/10 driver)
While creating a set of interrelated commerce entities at the outset of my scenario, I ran into an inability to set the `commerce_product_variation`'s base `price` field, and the `commerce_product`'s `variations` field.  I tracked it down to a combination of missing entity-level functionality in this and the related drupalextension project.

In this DrupalDriver project, I found that special "expanding" is done to field values prior to actual entity creation.  This is absolutely necessary for setting an entity reference field like `commerce_product`.`variations`.  Unfortunately, expansion is currently only done on non-base fields, preventing me from successfully setting a base field like `commerce_product`.`variations` from my scenario.

I also found this driver currently assumes an entity ID key of `id`, which is not universal to all entity types.  Going instead with whatever the entity type's ID key actually is would make entity clean up more straightforward and would open up entityCreate() to be utilized across node, user, and taxonomy term creation methods (reducing repeated entity creation code and centralizing where an entity, of whatever type, is created).

chrisolof:
Resolves #270.

Pairs really well with jhedstrom/drupalextension/pull/661.

### DIFF
--- a/src/Drupal/Driver/Cores/AbstractCore.php
+++ b/src/Drupal/Driver/Cores/AbstractCore.php
@@ -81,6 +81,9 @@ abstract class AbstractCore implements CoreInterface {
     $field_types = $this->getEntityFieldTypes($entity_type, $base_fields);
     foreach ($field_types as $field_name => $type) {
       if (isset($entity->$field_name)) {
+        // Consistently expand into an array of field values.
+        $entity->$field_name = !isset($entity->$field_name[0]) ? [$entity->$field_name] : $entity->$field_name;
+        // Expand array of field values using appropriate field handler.
         $entity->$field_name = $this->getFieldHandler($entity, $entity_type, $field_name)
           ->expand($entity->$field_name);
       }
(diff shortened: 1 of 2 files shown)


AlexSkrypnyk:
Hey @chrisolof - thanks for this contribution! This is a significant refactor that consolidates entity creation into a unified flow. We're planning to tackle this as part of the v3.x effort (see #312) where we'll have kernel tests to properly validate the changes. I'll leave this open as reference for that work. Appreciate the thoughtful implementation!

AlexSkrypnyk:
Hey @chrisolof - thanks for this contribution! Master has moved on significantly since this was opened (v3.x restructure, capability interfaces, kernel tests), so a clean rebase was not viable - but your insight absolutely was. I have opened #346 which reimplements the two bugs you identified against current master: auto-detection of base fields in expandEntityFields so commerce_product.variations and similar base entity-reference fields go through the handler pipeline, and dynamic id key resolution in entityCreate/entityDelete. Your original PR is attributed at the top of #346, and there is a dedicated kernel test covering the commerce_product to variations scenario you originally hit. I will close this PR once #346 lands. Appreciate you catching these!

AlexSkrypnyk:
Landed via #346. Thanks again, @chrisolof!

PR #273, opened 2024-03-11
Rely on project-level installed Drush version

### LINKED ISSUE #272: Drush Launcher is dead, the default Drush binary path should be the path of Drush binary installed at the project level
A quote from similar issues about the deprecation of Drush Launcher.

> Triggered by [the recent presentation](https://docs.google.com/presentation/d/14euH9Knf9Eaclw9PnVorfhxrFX4Ktph6ZdexBrWNWPc/edit#slide=id.g2161bd1e163_0_65) on DrupalCon Pittsburg.
> 
> **Drush 11 got an EOL date: 2023 November**. It was also marked as not recommended for Drupal 10. The recommended version is Drush 12.
> 
> Drush Launcher is not (officially) compatible with Drush 12 at this moment and it may not be compatible in the future, rather Drush 12 is going to fail nicely when it is called via Drush Launcher: https://github.com/drush-ops/drush-launcher/issues/105

mxr576:
Closes #272

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
   "require-dev": {
     "composer/installers": "^2.1",
     "drupal/coder": "~8.3.0",
-    "drush/drush": "^12",
+    "drush/drush": "^11 || ^12",
     "phpspec/phpspec": "~2.0 || ~4.0 || ~6.1 || dev-main",
     "phpunit/phpunit": "~6.0 || ~7.0 || ^9",
     "mockery/mockery": "^1.5",
--- a/composer.json
+++ b/composer.json
@@ -13,6 +13,7 @@
   ],
   "require": {
     "php": ">=7.4",
+    "composer-runtime-api": "^2.2.2",
     "symfony/process": "~2.5|~3.0|~4.4|^6",
     "symfony/dependency-injection": "~2.6|~3.0|~4.4|^6",
     "drupal/core-utility": "^8.4 || ^9 || ^10@beta"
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -13,13 +13,15 @@
...
     "php": ">=7.4",
+    "composer-runtime-api": "^2.2.2",
     "symfony/process": "~2.5|~3.0|~4.4|^6",
     "symfony/dependency-injection": "~2.6|~3.0|~4.4|^6",
     "drupal/core-utility": "^8.4 || ^9 || ^10@beta"
   },
   "require-dev": {
     "composer/installers": "^2.1",
     "drupal/coder": "~8.3.0",
+    "drush/drush": "^11 || ^12",
```
mxr576, line 24: Since only Drush >=12  is supported, Drush 11 should not be on the list.

https://www.drush.org/12.4.4/install/#drupal-compatibility

pcambra:
I think this could be an issue on environments such as Pantheon where they go a bit behind on drush versions and rely on the global drush still: https://docs.pantheon.io/guides/drush/drush-versions#changing-the-global-drush-version-on-pantheon

AlexSkrypnyk:
Hey @mxr576 - thanks for this contribution! I've opened #323 which builds on your work with a simpler approach - resolving via COMPOSER_BIN_DIR and vendor/bin/drush fallback. This complements the downstream fix in jhedstrom/drupalextension#799. Your contribution is attributed there. I'll close this PR once the new one is merged. Appreciate the work!

AlexSkrypnyk:
Landed via #323. Thanks again, @mxr576!

PR #276, opened 2024-07-29
D11 testing and fixes

Berdir:


Berdir:
Trying to support such a far range of versions is getting more and more challenging.

* Removed drupal-check, no longer really recommended and doesn't yet support symfony 7, phpstan should be setup directly, but no idea if that works for non-modules anyway, it certainly didn't find user_role_names(). Setting up phpstan seems out of scope.
* phpspec somehow doesn't seem to have a compatible version for PHP 8.1, so dropped combinations with that. In general, switched from exclude to including a few specific combinations only. too many invalid combinations now.
* phpunit 10 is necessary for D11, requires some changes to test classes and making the data provider abstract. Apart from the drush driver test, the only test is Drupal 7 field value stuff, but Drupal 7 isn't supported anymore.

jhedstrom:
Thank you!

PR #277, opened 2024-10-15
Allow D8 link handler to work with named field properties

### LINKED ISSUE #192: LinkHandler expects numeric indices
This code in `LinkHandler` expects numeric indices

```php
      $return[] = array(
        'options' => $options,
        'title' => $value[0],
        'uri' => $value[1],
      );
```
while the Drupal Extension actually allows passing in the proper entity properties:
```gherkin
| field_foo:uri   | http://example |
| field_foo:title | My link              |
```

chrisolof:
Resolves #192.

### DIFF
--- a/src/Drupal/Driver/Fields/Drupal8/LinkHandler.php
+++ b/src/Drupal/Driver/Fields/Drupal8/LinkHandler.php
@@ -11,21 +11,22 @@ class LinkHandler extends AbstractHandler {
    * {@inheritdoc}
    */
   public function expand($values) {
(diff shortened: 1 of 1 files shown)


AlexSkrypnyk:
Hey @chrisolof - thanks for this contribution! It's been a while since the last update here, so I've picked up the work and iterated on it. I've opened #311 which builds on your changes with unit tests for the new named property support. Your contribution is attributed there. I'll close this PR once the new one is merged. Appreciate the initial work!

PR #279, opened 2024-11-03
Avoid referencing UID 0 not on purpose

### LINKED ISSUE #278: When using a name for a user in an entity reference field, uid 0 gets chosen
I am using a fairly out of the box setup with creating some users and then using them as entity references.

An excerpt from my steps:

```
  Background:
    Given users:
      | name | roles | mail |
      # Some other stuff (...)
      | test_user_with_enterprise_team       |  | testteamenterprise@example.com |

    Given "team" content:
      | title | field_plan | field_team_members |
      | test team  | plan_enterprise | test_user_with_enterprise_team |
```

This has been working quite fine, up until upgrading to v.2.3.0.

Now, the entity reference handler that expands the field chooses uid 0 instead. Basically it translates to an SQL query like so:

```
SELECT "base_table"."uid" AS "uid", "base_table"."uid" AS "base_table_uid"
FROM
"users" "base_table"
LEFT JOIN "users_field_data" "users_field_data" ON "users_field_data"."uid" = "base_table"."uid"
WHERE ("users_field_data"."uid" = 'test_user_with_enterprise_team') or ("users_field_data"."name" LIKE 'test\\_user\\_with\\_enterprise\\_team' ESCAPE '\\')
```

To me this looks quite OK, but for some reason this gives me 2 hits. One is the correct one, but the first one is uid 0.

I can easily reproduce it directly in my sql, without any test setup:

```
mysql> select * from users LEFT JOIN users_field_data ON users_field_data.uid = users.uid WHERE users_field_data.uid = 'test_user_with_enterprise_team';
+-----+--------------------------------------+----------+------+----------+--------------------+--------------------------+------+------+------+----------+--------+------------+------------+--------+-------+------+------------------+
| uid | uuid                                 | langcode | uid  | langcode | preferred_langcode | preferred_admin_langcode | name | pass | mail | timezone | status | created    | changed    | access | login | init | default_langcode |
+-----+--------------------------------------+----------+------+----------+--------------------+--------------------------+------+------+------+----------+--------+------------+------------+--------+-------+------+------------------+
|   0 | 9656d6d9-060a-4fc4-b3f2-f8e131e17e53 | en       |    0 | en       | en                 | NULL                     |      | NULL | NULL | NULL     |      0 | 1730625944 | 1730625944 |      0 |     0 | NULL |                1 |
+-----+--------------------------------------+----------+------+----------+--------------------+--------------------------+------+------+------+----------+--------+------------+------------+--------+-------+------+------------------+
1 row in set, 1 warning (0.00 sec)
```

I am probably not an SQL expert enough to tell you why it works like that, but it surely is a regression, and it totally breaks many many of my tests :stuck_out_tongue: 

I see it was introduced with this https://github.com/jhedstrom/DrupalDriver/pull/241

To me this indicates we could probably add an additional condition on uid not being 0 at least.

eiriksm:
This fixes #278

### DIFF
--- a/src/Drupal/Driver/Fields/Drupal8/EntityReferenceHandler.php
+++ b/src/Drupal/Driver/Fields/Drupal8/EntityReferenceHandler.php
@@ -39,7 +39,10 @@ class EntityReferenceHandler extends AbstractHandler {
     foreach ((array) $values as $value) {
       $query = \Drupal::entityQuery($entity_type_id);
       $or = $query->orConditionGroup();
-      $or->condition($id_key, $value)
+      $id_group = $query->andConditionGroup()
+        ->condition($id_key, $value)
+        ->condition($id_key, 0, '<>');
+      $or->condition($id_group)
         ->condition($label_key, $value);
       $query->condition($or);
       $query->accessCheck(FALSE);


vaish:
This may work for MySQL /Mariadb but won't fix the issue for PostgreSQL as reported in https://github.com/jhedstrom/DrupalDriver/issues/278#issuecomment-2455989015.

AlexSkrypnyk:
Hey @eiriksm - thanks for this contribution! The underlying entity reference issue has been resolved by #305 which was merged recently. Appreciate you flagging the problem.

PR #282, opened 2024-11-15
Switch from "docker-compose" to "docker compose"

guillaume-a:
`docker-compose` is not available anymore so github workflow fails.
Replace with `docker compose` fixes the issue.


jhedstrom:
Thanks!

PR #284, opened 2025-01-15
Do not process empty batches.

idimopoulos:
Hi @jhedstrom 
I have had a weird issue with one of my behat test so before anything, this can be fixed by filtering users anew at https://github.com/jhedstrom/drupalextension/blob/main/src/Drupal/DrupalExtension/Context/RawDrupalContext.php#L266 before we call for the process batch.

I managed to reproduce it in behat by simply creating a basic test 
```
@api
Feature: Ensure user cancellation is not breaking behat.

  Scenario: User cancels the account
    Given users:
      | name | pass |
      | test | test |
    Given I delete the user "test"
```

What happens, is that the `cleanUsers` tries to clean users by calling `user_cancel`. In `/core/modules/user/user.module:614` of Drupal, a check occurs for whether the user actually exists and just throws an error that we are trying to delete a user that does not exist.

However, in `RawDrupalContext`, the `::processCallback` is called and the `Drupal8` driver just does
```
public function processBatch() {
    $this->validateDrupalSite();
    $batch =& batch_get();
    $batch['progressive'] = FALSE;
    batch_process();
  }
```

Because the batch from `batch_get` is returned empty, the progressive key is added to an empty array and the process is called. Then the batch fails with an error `Warning: Undefined array key "sets" in /var/www/html/web/core/includes/form.inc line 925` and the behat test partially succeeds. 

As said above, this can be fixed here with this PR, or in `RawDrupalContext` from drupalextension or in both.

### DIFF
--- a/src/Drupal/Driver/Cores/Drupal8.php
+++ b/src/Drupal/Driver/Cores/Drupal8.php
@@ -196,9 +196,10 @@ class Drupal8 extends AbstractCore implements CoreAuthenticationInterface {
    */
   public function processBatch() {
     $this->validateDrupalSite();
-    $batch =& batch_get();
-    $batch['progressive'] = FALSE;
-    batch_process();
+    if ($batch =& batch_get()) {
+      $batch['progressive'] = FALSE;
+      batch_process();
+    }
   }
 
   /**


idimopoulos:
Please, note, as I forgot to mention, that the dummy step that I created "I delete the user 'test'", simply does a`$user->delete()` but it is a fake step that represents ANY way that a user can be deleted while the test is running. It can be through the UI by the user themselves (with email verification) - or by an admin, or by some cron operation, or by some spam mechanism, or by any way while the test is running and is outside the scope of the behat API.

claudiu-cristea:
Makes sense

AlexSkrypnyk:
Hey @idimopoulos - thanks for this contribution! It's been a while since the last update here, so I've picked up the work and iterated on it. I've opened #310 which builds on your fix. Your contribution is attributed there. I'll close this PR once the new one is merged. Appreciate the initial work!

AlexSkrypnyk:
Landed via #310. Thanks again, @idimopoulos!

PR #292, opened 2025-09-23
291: Fix error in EntityReferenceHandler when  is FALSE.

### LINKED ISSUE #291: Error in EntityReferenceHandler when $label_key is FALSE
There is error `'' not found` when trying to create entity with reference on entity which doesn't have defined label key in it's entity type e.g. commerce_license.

To fix this we can try to add condition and modify query:
    foreach ((array) $values as $value) {
    $query = \Drupal::entityQuery($entity_type_id);
    if ($label_key) {
       $or = $query->orConditionGroup();
       $or->condition($id_key, $value)
          ->condition($label_key, $value);
       $query->condition($or);
    }
   else {
      $query->condition($id_key, $value);
   }
  $query->accessCheck(FALSE);
  if ($target_bundles && $target_bundle_key) {
      $query->condition($target_bundle_key, $target_bundles, 'IN');

khiminrm:
Fixes https://github.com/jhedstrom/DrupalDriver/issues/291

### DIFF
--- a/src/Drupal/Driver/Fields/Drupal8/EntityReferenceHandler.php
+++ b/src/Drupal/Driver/Fields/Drupal8/EntityReferenceHandler.php
@@ -38,10 +38,15 @@ class EntityReferenceHandler extends AbstractHandler {
 
     foreach ((array) $values as $value) {
       $query = \Drupal::entityQuery($entity_type_id);
-      $or = $query->orConditionGroup();
-      $or->condition($id_key, $value)
-        ->condition($label_key, $value);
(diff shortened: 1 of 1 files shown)


AlexSkrypnyk:
Hey @khiminrm - thanks for this contribution! It's been a while since the last update here, so I've picked up the work and iterated on it. I've opened #308 which re-implements your fix against the current codebase (which now includes the numeric ID guard from #305). Your contribution is attributed there. I'll close this PR once the new one is merged. Appreciate the initial work!

PR #305, opened 2026-03-19
PostgreSQL crashes when trying to look up an entity reference field t…

svenbergryen:
…hat's connected using a string/title (the query tries to select a string from an integer field)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved entity-reference field filtering: queries no longer attempt ID-based matches for non-numeric inputs, while still matching by label. This reduces incorrect matches and improves accuracy when selecting referenced items.


AlexSkrypnyk:
Thank you for this contribution, @svenbergryen! Clean fix that properly guards the ID-key condition for PostgreSQL compatibility. Merging now.

PR #357, opened 2026-04-27
Fixed 'FileHandler' regression for existing managed file lookup by filename.

AlexSkrypnyk:
## Summary

`FileHandler` in v2.5.0 (commit `951a75b`) was rewritten to always call `file_get_contents()` on the raw value, breaking every test suite that references an existing managed file by filename (e.g. `document.pdf`). Those callers received `Warning: file_get_contents(document.pdf): Failed to open stream: No such file or directory` because the bare filename was treated as a relative filesystem path rather than a reference to a pre-created managed file entity.

This fix restores compatibility with the v2.4.3 pattern while preserving the v2.5.0 ability to import brand-new files from disk. The handler now uses a filesystem-first, entity-fallback resolution order: if the value is a real path on disk it reads and imports the file; otherwise it looks up an existing managed file by numeric id or `filename` property; if neither resolves it throws a descriptive exception.

### Source

- `src/Drupal/Driver/Fields/Drupal8/FileHandler.php` - rewrote `expand()` to delegate to a new `resolveTargetId()` helper; extracted `createFileFromPath()` (v2.5.0 import path) and `loadExistingManagedFile()` (v2.4.3 entity-lookup path) as protected helpers.

### Tests

- `tests/Drupal/Tests/Driver/Fields/Drupal8/FileHandlerTest.php` - added tests for lookup-by-filename, lookup-by-numeric-id, filesystem-path-wins-over-shared-filename, and descriptive-exception-on-unresolved-value; refactored container helpers into `setContainer()`, `createFileRepositoryReturning()`, `createFileStorage()`, and `createFileEntity()` to support both `file.repository` and `entity_type.manager`/file-storage stubs.

### Changelog

- `CHANGELOG.md` - added an `[Unreleased]` section with a `Fixed` entry describing the v2.5.0 regression and the resolution strategy.

## Before / After

```
v2.5.0 (broken)
================

 value (any)
     |
     v
 file_get_contents($value)
     |
     +-- FALSE --> Exception "Error reading file X."
     |
     +-- ok  --> writeData() --> save() --> target_id


After this fix (Option C: filesystem-first, entity-fallback)
=============================================================

 value
     |
     v
 is_file($value)?
     |
     +-- YES --> createFileFromPath()
     |               |
     |               +-- read fails  --> Exception
     |               +-- write fails --> Exception
     |               +-- ok          --> target_id (new entity)
     |
     +-- NO  --> loadExistingManagedFile()
                     |
                     +-- ctype_digit? --> storage->load(id)
                     |                       |
                     |                       +-- found --> target_id
                     |                       +-- NULL  --> try filename
                     |
                     +-- storage->loadByProperties(['filename' => value])
                             |
                             +-- found --> target_id (reused entity)
                             |
                             +-- NULL  --> Exception
                                          "Could not resolve file value ..."
```

## Summary by CodeRabbit

* **Bug Fixes**
  * File handler now reuses existing managed files when referenced by filename or numeric ID, instead of failing with an error. File creation from disk paths remains available as a fallback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+### Fixed
+  * Fixed `FileHandler` regression introduced in 2.5.0 where referencing an existing managed file by filename failed with `Failed to open stream: No such file or directory`. The handler now reuses an existing managed file when the value matches a numeric file id or `filename` property, falling through to the v2.5.0 path-on-disk import only when the value resolves to a real file.
+
 ## [2.5.0]
 ### Added
   * [#306](https://github.com/jhedstrom/DrupalDriver/pull/306) Added `TimeHandler` for the `time_field` contrib module.
--- a/src/Drupal/Driver/Fields/Drupal8/FileHandler.php
+++ b/src/Drupal/Driver/Fields/Drupal8/FileHandler.php
@@ -12,32 +12,93 @@ class FileHandler extends AbstractHandler {
    */
   public function expand($values) {
     $return = [];
+
     foreach ((array) $values as $value) {
-      $file_path = is_array($value) ? $value['target_id'] ?? $value[0] : $value;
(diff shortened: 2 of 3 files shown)


AlexSkrypnyk:
@coderabbitai review

